PR #66, opened 2026-05-16
fix: migration stuck on step 2 + ECONNREFUSED race

aaf2tbz:
## Summary

Fixes two critical bugs in the v0.33.0 self-updater migration flow that cause the updater to hang/crash during the Beta 4 → Beta 5 upgrade path.

### Bug 1: Migration hangs forever on step 2 ("Preserving user data")

**Root cause:** `copy_dir_recursive` used `fs::read_dir` + `is_dir()` which follows symlinks. Wine prefixes contain a `dosdevices` directory with symlinks like `z: -> /` and `c: -> ../drive_c`. The recursive copy attempted to traverse into `z:` and copy the **entire root filesystem** into `/tmp`, never completing.

**Fix:**
- `copy_dir_recursive` now uses `fs::symlink_metadata` to detect symlinks before calling `is_dir()`, and recreates them with `std::os::unix::fs::symlink` instead of following
- New `preserve_selective` function skips `dosdevices`, `windows`, and `ProgramData` from the Wine prefix (these are disposable runtime files, not user data)
- `restore_user_data` recreates `dosdevices/c:` and `dosdevices/z:` symlinks after restore
- `remove_old_runtime` also switched to `symlink_metadata` for the same reason
- Sub-progress messages added within preserve step so users see activity

### Bug 2: ECONNREFUSED crashes migration overlay on fresh relaunch

**Root cause:** `update.py` kills the backend, installs the new app, then relaunches via `open`. The new MetalSharp starts, shows the migration overlay, and `MigrationView.vue` immediately calls `POST /update/migrate/start` on mount. But the Rust backend spawned by `RustBridge.start()` is still going through `waitForReady()` (polling with 200ms-1s delays). The frontend's `ipcRenderer.invoke("backend:request")` hits `ECONNREFUSED 127.0.0.1:9274` and the error propagates to the UI, crashing the migration overlay.

**Fix:** `startMigration()` now retries up to 20 times (10 seconds total) on `ECONNREFUSED` and timeout errors, showing "Waiting for backend to start..." in the meantime.

## Testing

- Verified symlink handling: Wine `dosdevices/z: -> /` no longer followed
- Preserved data flow: only user data (Steam install, games, library) copied, not disposable Wine runtime files
- ECONNREFUSED retry: frontend waits for backend instead of crashing

### DIFF
--- a/app/src-rust/src/migrate.rs
+++ b/app/src-rust/src/migrate.rs
@@ -213,13 +213,16 @@ fn preserve_user_data(ms_dir: &PathBuf) -> PreservedData {
 
     let setup_json = ms_dir.join("setup.json").exists().then(|| fs::read(ms_dir.join("setup.json")).ok()).flatten();
 
+    write_migrate_progress("running", 2, 5, "Preserving user data (Steam prefix)...", None);
     let prefix_steam_tmp = tmp.join("prefix-steam");
     let prefix_steam = ms_dir.join("prefix-steam");
     if prefix_steam.exists() {
         let _ = fs::create_dir_all(&prefix_steam_tmp);
-        copy_dir_recursive(&prefix_steam, &prefix_steam_tmp);
+        let skip = ["dosdevices", "windows", "ProgramData"];
+        preserve_selective(&prefix_steam, &prefix_steam_tmp, &skip);
     }
 
+    write_migrate_progress("running", 2, 5, "Preserving user data (games)...", None);
     let games_tmp = tmp.join("games");
     let games = ms_dir.join("games");
     if games.exists() {
@@ -227,6 +230,7 @@ fn preserve_user_data(ms_dir: &PathBuf) -> PreservedData {
         copy_dir_recursive(&games, &games_tmp);
     }
 
+    write_migrate_progress("running", 2, 5, "Preserving user data (library)...", None);
     let sharp_library_tmp = tmp.join("sharp-library");
     let sharp_library = ms_dir.join("sharp-library");
     if sharp_library.exists() {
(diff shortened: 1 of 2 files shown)


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm taking a look at the fixes in `fix: migration stuck on step 2 + ECONNREFUSED race` (commit `bece6cf3`) and will follow up shortly.

app/src-rust/src/migrate.rs
```diff
@@ -274,6 +311,19 @@ fn restore_user_data(ms_dir: &PathBuf, preserved: &PreservedData) {
             let _ = fs::create_dir_all(&dst);
         }
         copy_dir_recursive(&preserved.prefix_steam_tmp, &dst);
```
PR-Reviewer-Ant, line 313: This restore copies the selectively preserved prefix back into the existing `prefix-steam` directory, so the skipped `dosdevices`, `windows`, and `ProgramData` directories are still left in place from the old prefix. That means the PR does not actually achieve the described cleanup of disposable Wine-prefix runtime files; it only avoids copying those directories into the temp backup. If those old prefix files are part of what Beta 4 to Beta 5 needs to refresh, this migration can complete while still leaving the user on stale prefix state. The restore path should either replace/clear the disposable subtrees before copying user data back, or build a fresh prefix destination and then recreate only the intended links/data.

app/src/renderer/components/MigrationView.vue
```diff
@@ -34,8 +37,14 @@ async function startMigration() {
...
     }
   } catch (e: any) {
-    error.value = e.message ?? "Network error";
-    message.value = `Error: ${error.value}`;
+    if (retriesLeft > 0 && (e.message?.includes("ECONNREFUSED") || e.message?.includes("timeout"))) {
+      message.value = "Waiting for backend to start...";
+      await new Promise((r) => setTimeout(r, START_RETRY_DELAY_MS));
+      startMigration(retriesLeft - 1);
+    } else {
+      error.value = e.message ?? "Network error";
```
PR-Reviewer-Ant, line 45: The retry call is fire-and-forget. After the 500ms delay, the current `startMigration` invocation resolves while a new async attempt continues outside the caller's lifecycle, so `onMounted` no longer tracks it and `onUnmounted` cannot cancel it. In practice this can still start polling or mutate component refs after the migration view has been torn down. Returning or awaiting the recursive retry would at least keep the retry chain tied to the original call; a mounted/cancel flag would make the lifecycle behavior explicit.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `bece6cf3`

</details>

The symlink-copy fix itself is pointed in the right direction, but the selective Wine-prefix preservation does not actually replace the old disposable prefix contents. I also found a smaller retry lifecycle issue in the migration overlay.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description says prefix subtrees such as dosdevices, windows, and ProgramData are skipped and then recreated, but restore_user_data copies the preserved data back into an existing prefix-steam directory without clearing those old subtrees. The Vue retry issue is directly visible in the changed catch block.

> **Note:** This PR touches UI files but no screenshots were referenced in the description. Consider adding visual previews for reviewers.

PR #74, opened 2026-05-16
ci: package required runtime assets

aaf2tbz:
## Summary
- package metalsharp_bundle2 and configs into the Electron app resources
- make the bundle downloader fetch the exact assets the setup wizard depends on
- make CI fail if required DMG install assets are missing instead of building without them
- repacked the bundles release metalsharp_bundle2 asset to include Goldberg, EAC toggle, Mono ARM64, DXMT add-ons, shims, and shader-cache content

## Asset audit
Required bundled assets now validated by CI:
- metalsharp_bundle.tar.zst
- metalsharp_bundle2.tar.zst
- SteamSetup.exe
- steamwebhelper.exe
- steamwebhelper-wrapper.c

Steam-based assets are preserved. Separate fallback assets on the bundles release were left in place because the installer can still download them individually.

## Validation
- node parsed app/package.json
- tools/dmg/create-bundles.sh downloaded required assets
- verified metalsharp_bundle2 contains goldberg/x86/steam_api.dll, goldberg/x64/steam_api64.dll, eac-toggle/x86_64-windows/_winhttp.dll, mono-arm64/bin/mono, DXMT d3d12.dll, and shims/libsteam_api.dylib
- npm run build
- npx electron-builder --dir --mac --publish never
- verified packaged app resources contain both bundles, Steam assets, and configs
- git diff --check

### DIFF
--- a/tools/dmg/create-bundles.sh
+++ b/tools/dmg/create-bundles.sh
@@ -15,7 +15,10 @@ mkdir -p "$BUNDLE_DIR"
 
 BUNDLES=(
     "metalsharp_bundle.tar.zst"
+    "metalsharp_bundle2.tar.zst"
     "SteamSetup.exe"
+    "steamwebhelper.exe"
+    "steamwebhelper-wrapper.c"
 )
 
 for bundle in "${BUNDLES[@]}"; do
@@ -25,12 +28,14 @@ for bundle in "${BUNDLES[@]}"; do
         continue
     fi
     echo "Downloading $bundle..."
-    curl -sL -o "$dest" "https://github.com/$REPO/releases/download/$RELEASE_TAG/$bundle"
-    if [ -f "$dest" ]; then
+    curl -fL -o "$dest" "https://github.com/$REPO/releases/download/$RELEASE_TAG/$bundle"
+    if [ -s "$dest" ]; then
         size=$(du -sh "$dest" | cut -f1)
         echo "  -> $dest ($size)"
     else
         echo "  FAILED: $bundle"
+        rm -f "$dest"
+        exit 1
     fi
 done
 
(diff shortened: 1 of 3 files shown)


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm taking a look at the changes in `ci: package required runtime assets` (commit `e32af994`) and will follow up shortly.

tools/dmg/create-bundles.sh
```diff
@@ -15,7 +15,10 @@ mkdir -p "$BUNDLE_DIR"
 
 BUNDLES=(
     "metalsharp_bundle.tar.zst"
+    "metalsharp_bundle2.tar.zst"
     "SteamSetup.exe"
+    "steamwebhelper.exe"
+    "steamwebhelper-wrapper.c"
 )
```
PR-Reviewer-Ant, line 22: This skip condition accepts any existing file, including the zero-byte placeholders that the old `npm run dist` script used to create for `metalsharp_bundle.tar.zst`, `SteamSetup.exe`, and `steamwebhelper-wrapper.c`. In CI the following `test -s` loop catches that, but local `npm run dist` now runs this downloader and then packages whatever is present, so an existing empty or partial bundle can still be shipped. Please make the downloader validate `-s` before skipping, or remove/refetch invalid existing assets; ideally reuse the same required-asset check from CI for the npm packaging path too.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e32af994`

</details>

I found one packaging correctness issue: the downloader now treats any existing bundle path as valid, which can preserve the zero-byte placeholders that the old `dist` script created. That undermines the PR's stated goal of packaging the required runtime assets reliably.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed downloader explicitly skips on `-f` in tools/dmg/create-bundles.sh, while the removed package.json dist logic previously created empty placeholder files for several of the same bundle names. CI now catches this after the script, but `npm run dist` does not run the same required-asset validation before packaging.

PR #75, opened 2026-05-16
ci: reject empty bundle placeholders

aaf2tbz:
## Summary
- fixes the review comment from #74 about zero-byte bundle placeholders surviving local packaging
- skips only non-empty existing bundle assets
- removes and re-downloads empty/invalid existing bundle paths
- validates all required bundle assets at the end of the downloader, so CI and local npm packaging use the same guard

## Review context
Addresses #74 review thread: https://github.com/aaf2tbz/metalsharp/pull/74#discussion_r3253065826

## Validation
- bash -n tools/dmg/create-bundles.sh
- git diff --check
- simulated a zero-byte steamwebhelper-wrapper.c placeholder and verified create-bundles.sh removed and re-downloaded it
- reran create-bundles.sh on valid assets and verified it skips non-empty files

### DIFF
--- a/tools/dmg/create-bundles.sh
+++ b/tools/dmg/create-bundles.sh
@@ -23,12 +23,12 @@ BUNDLES=(
 
 for bundle in "${BUNDLES[@]}"; do
     dest="$BUNDLE_DIR/$bundle"
-    if [ -s "$dest" ]; then
+    if [ -f "$dest" ] && [ -s "$dest" ]; then
         echo "SKIP: $bundle — already exists"
         continue
     elif [ -e "$dest" ]; then
         echo "Invalid existing $bundle — removing and downloading fresh copy"
-        rm -f "$dest"
+        rm -rf "$dest"
     fi
     echo "Downloading $bundle..."
     curl -fL -o "$dest" "https://github.com/$REPO/releases/download/$RELEASE_TAG/$bundle"
@@ -43,7 +43,7 @@ for bundle in "${BUNDLES[@]}"; do
 done
 
 for bundle in "${BUNDLES[@]}"; do
-    if [ ! -s "$BUNDLE_DIR/$bundle" ]; then
+    if [ ! -f "$BUNDLE_DIR/$bundle" ] || [ ! -s "$BUNDLE_DIR/$bundle" ]; then
         echo "Missing required bundle: $bundle"
         exit 1
     fi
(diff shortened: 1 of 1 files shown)


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm taking a look at the changes in `ci: reject empty bundle placeholders` (commit `c2498b38`) and will follow up shortly.

tools/dmg/create-bundles.sh
```diff
@@ -23,9 +23,12 @@ BUNDLES=(
 
 for bundle in "${BUNDLES[@]}"; do
     dest="$BUNDLE_DIR/$bundle"
-    if [ -f "$dest" ]; then
+    if [ -s "$dest" ]; then
```
PR-Reviewer-Ant: `-s` does not prove this is a usable bundle file. A directory or other non-regular filesystem entry at `app/bundles/$bundle` can satisfy this check, get skipped as "already exists", and then satisfy the final validation too. That leaves local packaging with a broken bundle path even though this PR is meant to make CI and local packaging reject invalid placeholders. Please keep the non-empty check, but also require a regular file here and in the final validation, for example `if [ -f "$dest" ] && [ -s "$dest" ]; then ...` and treat any existing non-regular path as invalid.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `c2498b38`

</details>

The zero-byte placeholder case is handled, but the new guard accidentally accepts non-regular paths as valid bundle assets. That weakens the downloader’s final validation and can still let packaging proceed with a broken bundle entry.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The behavior follows directly from the changed shell predicates in tools/dmg/create-bundles.sh: the old regular-file check was replaced with -s, and the final validation also uses only -s.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `9e1df21a`

</details>

The update addresses the prior placeholder issue: empty files and non-regular existing paths are removed, re-downloaded, and the final guard now requires each bundle path to be a non-empty regular file. I did not find a new correctness, security, or data-integrity issue in this diff.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR changes only tools/dmg/create-bundles.sh, and the relevant control flow is fully visible in the diff. The previous concern about directories or other non-file paths passing the -s check is addressed by the new -f checks before skipping and in final validation.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #81, opened 2026-05-16
[codex] Make M12 the primary D3D Metal path

aaf2tbz:
## Summary
- make M12 the first stable DXMT D3D pipeline and the unresolved default
- make D3D12 PE detection win before broad directory heuristics that could push games back to M11
- strengthen native D3D12 Metal coverage with real compute PSO dispatch, offscreen indexed draw, C entrypoint coverage, and an M12 pipeline map

## Validation
- cargo fmt --all -- --check
- cargo test
- cargo clippy --all-targets -- -D warnings
- cmake --build build
- ctest --test-dir build --output-on-failure
- npm run build
- npx biome check src/
- git diff --check
- git diff --cached --check


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm taking a look at the changes in `[codex] Make M12 the primary D3D Metal path` (commit `d085c863`) and will follow up shortly.

src/d3d/d3d12/D3D12CommandList.mm
```diff
@@ -664,7 +729,7 @@ void execute(MetalDevice& metalDev) {
                     }
 
                     if (m_rootSignature && !m_argumentBuffer.empty()) {
-                        id<MTLDevice> mtlDev = MTLCreateSystemDefaultDevice();
+                        id<MTLDevice> mtlDev = (__bridge id<MTLDevice>)metalDev.nativeDevice();
                         if (mtlDev) {
                             id<MTLBuffer> argBuf = [mtlDev newBufferWithBytes:m_argumentBuffer.data()
                                                                        length:m_rootSignature->argumentBufferSize
```
PR-Reviewer-Ant, line 734: Vertex buffer binding is gated on `m_rootSignature && !m_argumentBuffer.empty()`, so draws that use plain input assembly without a root signature never bind the buffers recorded by `IASetVertexBuffers`. The new indexed MSL test does exactly that: it creates a PSO and calls `IASetVertexBuffers`, but never calls `SetGraphicsRootSignature`, while the shader reads `[[buffer(0)]]`. Binding vertex/index buffers should not depend on whether an argument buffer/root signature is active.

include/metalsharp/D3D12Device.h
```diff
@@ -566,6 +570,7 @@ class D3D12DeviceImpl final : public ID3D12Device {
                 std::unique_ptr<MetalBuffer>(MetalBuffer::create(*m_metalDevice, (size_t)pDesc->Width, nullptr));
             if (res->metalBuffer) {
                 res->m_gpuAddress = reinterpret_cast<UINT64>(res->metalBuffer->nativeBuffer());
+                m_gpuAddressResources[res->m_gpuAddress] = res;
```
PR-Reviewer-Ant: `m_gpuAddressResources` stores raw `D3D12ResourceImpl*` entries when buffers are created, but resources can later be released and deleted without removing their entry from this map. A later `findResourceForGPUAddress()` can dereference a dangling pointer during vertex/index binding, which is a use-after-free path and can bind the wrong Metal buffer or crash. The device needs ownership/refcounting or unregister-on-destroy semantics for this lookup table.

docs/m12-pipeline-map.md
```diff
@@ -0,0 +1,105 @@
...
+3. M12 deploys DXMT PE DLLs into the game directory:
+   `d3d12.dll`, `d3d11.dll`, `dxgi.dll`, `d3d10core.dll`, and `winemetal.dll`.
+4. M12 sets `WINEDLLOVERRIDES` so Wine prefers the deployed native DXMT DLLs.
+5. M12 adds DXMT/Wine unix library paths to `DYLD_FALLBACK_LIBRARY_PATH`.
+6. M12 sets shader and pipeline cache paths under the MetalSharp cache root.
+7. Wine launches the executable with the M12 launch args, including `-dx12`.
+8. DXMT handles D3D12/DXGI calls, compiles DXIL/MSL work, sends commands through
+   `winemetal`, and presents through the Wine/macOS surface.
+
+## Current Verification
```
PR-Reviewer-Ant, line 36: This new public doc includes local machine-specific paths like `/Users/alexmondello/repos/metalsharp` and `/Volumes/AverySSD/metalsharp/dxmt-src`. The verification narrative is useful, but these should be generalized or moved into private notes so the repo docs do not bake in one maintainer's filesystem layout as source evidence.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `d085c863`

</details>

I found two correctness issues in the new native D3D12 path: vertex/index buffers are only bound when a root signature is present, and GPU-address lookup keeps raw resource pointers after resources can be released. There is also a docs leak of local machine paths in the new M12 map.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The issues are visible directly in the changed files: IASetVertexBuffers records Metal buffers, but execute only binds them inside the root-signature argument-buffer branch; CreateCommittedResource inserts raw D3D12ResourceImpl pointers into m_gpuAddressResources with no corresponding removal in Release/destruction. The docs path leak is explicit in docs/m12-pipeline-map.md.

aaf2tbz:
Addressed the three M12 review threads in `Address M12 native review findings`: vertex buffers recorded by `IASetVertexBuffers` are now bound independently of root-signature argument buffers, GPU-address resources are retained by the device lookup table until device teardown, and `docs/m12-pipeline-map.md` no longer includes local maintainer filesystem paths. Added `test_d3d12_gpu_address_lookup_retains_resource` to cover the resource-lifetime path. Local checks: `cmake --build build --target test_d3d12 test_d3d12_entrypoint test_phase18`, `./build/tests/test_d3d12`, `./build/tests/test_d3d12_entrypoint`, `./build/tests/test_phase18`, `ctest --test-dir build -R d3d12|d3d12_entrypoint|phase18|phase19 --output-on-failure`, `cargo fmt --all -- --check`, `cargo test`, `cargo clippy --all-targets -- -D warnings`, `npm run build`, `npx biome check src/`, `git diff --check`.

src/d3d/d3d12/D3D12CommandList.mm
```diff
@@ -901,6 +1002,70 @@ void execute(MetalDevice& metalDev) {
...
+    }
+    if (!ok) {
+        ok = m_shaderTranslator->translateDXBC(shaderData, shaderSize, ShaderStage::Compute, compiled);
+    }
+
+    id<MTLFunction> function = nil;
+    if (ok && compiled.computeFunction) {
+        function = (__bridge id<MTLFunction>)compiled.computeFunction;
+    } else if (looksLikeMSL(shaderData, shaderSize)) {
+        id<MTLDevice> mtlDevice = (__bridge id<MTLDevice>)m_metalDevice->nativeDevice();
```
PR-Reviewer-Ant, line 1040: On failure this leaves `*ppPSO` pointing at a partially initialized pipeline object while returning `E_FAIL`. That breaks the usual COM/D3D creation contract: callers commonly only release/use the out pointer on `SUCCEEDED(hr)`, so this can leak the object and, worse, leaves a non-null failed PSO that later code may accidentally treat as usable. Please only assign `*ppPSO` after the Metal compute pipeline is created successfully, or clear/release it before returning failure.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `6510c451`

</details>

I found one D3D12 COM-contract issue in the new compute PSO path. The routing/default changes otherwise line up with the PR description.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The issue is visible directly in the new CreateComputePipelineState implementation: it writes a newly allocated object to ppPSO before shader compilation/pipeline creation can fail, then returns E_FAIL without clearing that output. I did not run the Metal runtime tests, so confidence is limited to diff-level behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

aaf2tbz:
Addressed the latest compute PSO COM-contract review in `Fix compute PSO failure output`: `CreateComputePipelineState` now initializes `*ppPSO` to null, only assigns a PSO on success or intentional empty-PSO success cases, and releases/clears the partial object before returning `E_FAIL`. Added a D3D12 test asserting failed compute PSO creation leaves the output null. Local checks: `cmake --build build --target test_d3d12`, `./build/tests/test_d3d12`, `cmake --build build --target test_d3d12_entrypoint test_phase18`, `./build/tests/test_d3d12_entrypoint`, `./build/tests/test_phase18`, `ctest --test-dir build -R d3d12|d3d12_entrypoint|phase18|phase19 --output-on-failure`, `cargo fmt --all -- --check`, `cargo test`, `cargo clippy --all-targets -- -D warnings`, `npm run build`, `npx biome check src/`, `git diff --check`.

include/metalsharp/D3D12Device.h
```diff
@@ -566,6 +578,8 @@ class D3D12DeviceImpl final : public ID3D12Device {
                 std::unique_ptr<MetalBuffer>(MetalBuffer::create(*m_metalDevice, (size_t)pDesc->Width, nullptr));
```
PR-Reviewer-Ant, line 588: This permanently extends the lifetime of every buffer resource created on the device. `CreateCommittedResource` now `AddRef`s each buffer and stores it in `m_gpuAddressResources`, but entries are only released from `D3D12DeviceImpl::~D3D12DeviceImpl`. Real D3D12 workloads create and release lots of transient upload/default buffers; with this design, released buffers stay alive until device teardown, so VRAM/shared memory usage can grow without bound during a game session. The lookup table needs either a deregistration path when the resource's public refcount reaches zero, or a non-owning/weak registration scheme plus command-list retention only for resources actually bound during recording.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `36581c5b`

</details>

I found one introduced lifetime bug in the native D3D12 path: buffer resources are now retained by the device forever for GPU-address lookup. That makes transient D3D12 buffers leak until device teardown and can grow badly under real game workloads.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The diff adds an AddRef and inserts every buffer resource into m_gpuAddressResources in CreateCommittedResource, while the only release path for those retained resources is D3D12DeviceImpl::~D3D12DeviceImpl. The added phase18 test also explicitly releases the resource and expects lookup/binding to keep working afterward, confirming the lifetime behavior is intentional in this patch.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

aaf2tbz:
Addressed the latest GPU-address lifetime review in `Use weak D3D12 GPU address registry`: the device lookup table no longer owns every buffer until teardown. Buffer resources now unregister themselves from the GPU-address map when their public refcount reaches zero, and command lists retain only resources actually bound through GPU-address lookup while recording. Added tests for both unregister-on-release and command-list retention of a bound resource. Local checks: `cmake --build build --target test_phase18 test_d3d12`, `./build/tests/test_phase18`, `./build/tests/test_d3d12`, `ctest --test-dir build -R d3d12|d3d12_entrypoint|phase18|phase19 --output-on-failure`, `cargo fmt --all -- --check`, `cargo test`, `cargo clippy --all-targets -- -D warnings`, `npm run build`, `npx biome check src/`, `git diff --check`, plus local reproduction of the PR C/C++ job via `cmake --build build-native --parallel $(sysctl -n hw.ncpu)` and `ctest --test-dir build-native --output-on-failure -E ^(metal_device|dxbc|format_translation|phase17)$`.

src/d3d/d3d12/D3D12CommandList.mm
```diff
@@ -203,19 +252,28 @@ HRESULT IASetVertexBuffers(UINT start, UINT num, const D3D12_VERTEX_BUFFER_VIEW*
...
             }
             return S_OK;
         }
 
         HRESULT IASetIndexBuffer(const D3D12_INDEX_BUFFER_VIEW* pView) override {
             if (!pView)
                 return S_OK;
-            m_indexBuffer.offset = pView->BufferLocation;
+            UINT64 offset = 0;
+            D3D12ResourceImpl* res = device.findResourceForGPUAddress(pView->BufferLocation, &offset);
```
PR-Reviewer-Ant, line 270: If `findResourceForGPUAddress()` misses, this stores the D3D12 GPU virtual address as though it were an Objective-C `MTLBuffer` object. The execute path later bridges `m_indexBuffer.metalBuffer` to `id<MTLBuffer>` and passes it to Metal, so any stale, zero, out-of-range, or externally supplied GPU address can become an invalid Objective-C object dereference. The vertex-buffer path has the same fallback. A miss should leave the binding null, fail the call, or otherwise avoid treating an opaque D3D12 address as a Metal object.

include/metalsharp/D3D12Device.h
```diff
@@ -438,8 +441,16 @@ class D3D12ResourceImpl final : public ID3D12Resource {
     std::unique_ptr<MetalTexture> metalTexture;
     UINT m_resourceState;
     UINT64 m_gpuAddress = 0;
+    std::unordered_map<UINT64, D3D12ResourceImpl*>* m_gpuAddressRegistry = nullptr;
 
     D3D12ResourceImpl(const D3D12_RESOURCE_DESC& d) : desc(d), m_resourceState(D3D12_RESOURCE_STATE_COMMON) {}
```
PR-Reviewer-Ant, line 450: The resource now keeps a raw pointer to `D3D12DeviceImpl::m_gpuAddressResources` and erases from it in the resource destructor. That is only safe while the device object is still alive, but COM resources can outlive the device that created them unless the resource holds a device reference. Releasing the device before releasing a buffer leaves `m_gpuAddressRegistry` dangling and the resource destructor will write through freed device storage.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `7fac18aa`

</details>

I found two correctness issues in the new native D3D12 GPU-address plumbing. Both are reachable from the changed command-list/resource lifetime paths and can turn ordinary fallback cases into crashes or use-after-free behavior.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed D3D12 command-list code explicitly casts unresolved D3D12 GPU virtual addresses to Metal buffer objects, and the changed resource destructor stores a raw pointer back into the owning device's registry. Both behaviors are visible in the provided changed files and do not require guessing about unrelated modules.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

aaf2tbz:
Fixed the PR CI C/C++/Obj-C format failure in `Format M12 phase18 test`. The failing gate was clang-format on `tests/test_phase18.cpp`; local validation now passes the same format command: `find src include tests tools \( -name '*.c' -o -name '*.cpp' -o -name '*.h' -o -name '*.mm' \) -print0 | xargs -0 clang-format --dry-run --Werror`. Also reran `cmake --build build --target test_phase18`, `./build/tests/test_phase18`, and `git diff --check`.

include/metalsharp/D3D12Device.h
PR-Reviewer-Ant, line 450: This destructor dereferences `m_gpuAddressRegistry`, which points into the owning `D3D12DeviceImpl`. Nothing in this patch makes resources retain the device or clears the registry pointer when the device is destroyed, so a resource that outlives its device will erase from freed memory on `Release()`. D3D12 resources commonly outlive local device variables in tests and callers; the registry needs ownership/lifetime protection rather than a raw pointer back into the device.

src/d3d/d3d12/D3D12CommandList.mm
```diff
@@ -203,19 +252,28 @@ HRESULT IASetVertexBuffers(UINT start, UINT num, const D3D12_VERTEX_BUFFER_VIEW*
             if (!pViews)
                 return S_OK;
             for (UINT i = 0; i < num && (start + i) < 32; ++i) {
-                m_vertexBuffers[start + i].offset = pViews[i].BufferLocation;
+                UINT64 offset = 0;
+                D3D12ResourceImpl* res = device.findResourceForGPUAddress(pViews[i].BufferLocation, &offset);
+                m_vertexBuffers[start + i].metalBuffer =
```
PR-Reviewer-Ant: When a GPU virtual address is not found in the registry, this falls back to treating the address value as an Objective-C `MTLBuffer*`. A stale resource, foreign address, or address from a resource released before binding will later be bridged and passed to `setVertexBuffer`, which can turn a normal D3D12 misuse or lifetime edge into a native crash. For the new registry path, unresolved GPU addresses should fail/skip deterministically instead of being cast into Metal object pointers.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `3b82dbe7`

</details>

I found two lifetime hazards in the new D3D12 GPU-address tracking path. Both come from storing raw resource/device pointers without tying their lifetimes together.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new registry is introduced directly in D3D12DeviceImpl and D3D12ResourceImpl, and the changed command-list binding code consumes it. The lifetime failure modes are visible from the changed lines without needing runtime repro.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `0ebd9397`

</details>

I did not find a concrete correctness, security, data-integrity, or breaking-change issue in the changed lines. The implementation matches the PR description at the routing level: M12 is now primary/stable, D3D12 PE detection runs before broad directory heuristics, and the native D3D12 tests/entrypoint coverage were expanded.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed Rust routing and pipeline definitions directly support the stated M12 behavior, and the C++ changes include targeted tests for compute PSO creation, indexed draw binding, GPU-address lifetime, and the D3D12 C entrypoint. I could not independently exercise a real Wine-launched M12 game/runtime DLL deployment path from the provided context, so runtime confidence is limited to the diff and the validation listed in the PR description.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

src/d3d/d3d12/D3D12CommandList.mm
```diff
@@ -858,13 +939,24 @@ void execute(MetalDevice& metalDev) {
         PipelineStateDesc pipeDesc;
         pipeDesc.vertexStride = 0;
 
+        for (UINT i = 0; pDesc->InputLayout && i < pDesc->NumInputElements; ++i) {
+            const D3D12_INPUT_ELEMENT_DESC& elem = pDesc->InputLayout[i];
```
PR-Reviewer-Ant, line 943: This stride calculation treats `AlignedByteOffset` as a literal byte offset. In D3D12, input elements commonly use `D3D12_APPEND_ALIGNED_ELEMENT` (`0xffffffff`) to ask the runtime to append after the previous element. With the new code, that value makes `elemEnd` wrap or become huge, so `pipeDesc.vertexStride` can be wrong and the Metal pipeline will bind vertex buffers with an invalid stride. Please handle the append-aligned sentinel by tracking the running offset per input slot before computing `vertexStride`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `3c51d2ba`

</details>

I found one correctness issue in the new native D3D12 pipeline-state path: append-aligned input elements can produce a bogus vertex stride. The MTSP routing and version/documentation updates otherwise line up with the PR description from the diff provided.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The changed CreateGraphicsPipelineState code now derives vertexStride from D3D12_INPUT_ELEMENT_DESC::AlignedByteOffset directly, and the failure mode is visible in the diff. I did not have a runtime repro or full build output, so I am limiting this to the concrete D3D12 input-layout handling issue.


**Since my last review:**
- Fixed or likely addressed: 0
- Still blocking: 0
- Newly found: 1

PR #84, opened 2026-05-17
[codex] Fix DOOM Eternal Steam launch

aaf2tbz:
## Summary
- route DOOM and DOOM Eternal to the Wine Steam pipeline instead of DXMT M11/M12 presets
- persist a DOOM Eternal Steam launch-option override that bypasses the invisible idTech launcher and starts `DOOMEternalx64vk.exe` through `%COMMAND%`
- keep Wine Steam running on subsequent Steam launches, killing only known DOOM game-side processes before relaunch
- update the tracked `app/src/main/metalsharp-backend` binary from the rebuilt release backend

## Root Cause
DOOM Eternal's current Steam config launches `launcher/idTechLauncher.exe`, which can remain as an invisible Wine window. The app was also passing cache env through the Steam pipeline, which caused Wine Steam to be stopped and relaunched before every game launch. That made the Electron request path prone to socket timeouts while Steam was restarting.

## Validation
- `cargo test` (32 passed)
- `cargo build --release`
- `git diff --cached --check`

### DIFF
--- a/app/src-rust/src/launch.rs
+++ b/app/src-rust/src/launch.rs
@@ -52,26 +52,21 @@ pub fn launch_via_steam_with_env(
     let prefix = home.join(".metalsharp").join("prefix-steam");
     let prefix_str = prefix.to_string_lossy().to_string();
 
-    if extra_env.is_empty() {
-        if !crate::steam::is_wine_steam_running() {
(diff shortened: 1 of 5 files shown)


aaf2tbz:
Closing this draft PR per owner direction; this DOOM Eternal branch is not being carried forward.

PR #89, opened 2026-05-18
feat(d3d12): expanded format coverage, root signature serialization, fence sync, HDR swapchain

aaf2tbz:
## Summary

Alignment of MetalSharp's D3D12/DXGI translation layer with full format coverage and missing entry points.

- **Format coverage**: DXGI format table expanded from 16 to 100+ formats — full typeless, sRGB, depth, BC1-7, HDR, and packed formats with Metal pixel format mapping
- **Root signature serialization**: `D3D12SerializeRootSignature` and `D3D12SerializeVersionedRootSignature` — generates DXBC-compatible binary blobs from root signature descriptors
- **Entry point stubs**: `D3D12GetDebugInterface` (returns `E_NOINTERFACE`), `D3D12EnableExperimentalFeatures` (pass-through)
- **Fence sync**: `os_sync_wait_on_address` + `os_unfair_lock` + `os_sync_wake_by_address_all` — efficient CPU-GPU fence waiting
- **HDR swapchain**: Format-aware `CAMetalLayer` pixel format selection, `wantsExtendedDynamicRangeContent`, extended linear sRGB colorspace for RGBA16Float/XR_BIAS formats

## Files changed (9, +547/-45)

| File | Change |
|---|---|
| `include/metalsharp/FormatTranslation.h` | 100+ DXGI format enums |
| `src/metal/Sampler.mm` | Full `dxgiFormatToMetal()`, `dxgiFormatToBytesPerPixel()`, `dxgiFormatIsDepth()` |
| `src/d3d/d3d12/EntryPoint.cpp` | Root signature serialization + debug stubs |
| `include/metalsharp/D3D12Device.h` | `os_sync_wait_on_address` fence |
| `src/dxgi/DXGISwapChain.mm` | HDR swapchain with colorspace |
| `include/metalsharp/DXGI.h` | Updated signatures for format passthrough |
| `include/d3d/D3D12.h` | Added `DXGI_FORMAT_R10G10B10_XR_BIAS_A2_UNORM` |
| `include/d3d/D3D11.h` | Added `DXGI_FORMAT_R10G10B10_XR_BIAS_A2_UNORM` |
| `tests/test_format_translation.cpp` | Updated for new format aliases |

## Test plan

- [x] Clean build (0 errors, 0 warnings)
- [x] `test_format_translation` passes
- [ ] Rain World smoke test via MetalSharp Wine launcher
- [ ] Steam CEF swapchain format negotiation
- [ ] HDR display output verification


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm taking a look at the feature work in `feat(d3d12): expanded format coverage, root signature serialization, fence sync, HDR swapchain` (commit `fec5d129`) and will follow up shortly.

src/d3d/d3d12/EntryPoint.cpp
```diff
@@ -18,4 +21,100 @@ HRESULT D3D12CreateDevice(void* pAdapter, UINT MinimumFeatureLevel, const GUID&
...
     return hr;
 }
+
+static void writeRootSignatureHeader(uint8_t* out, uint32_t numParameters, uint32_t numStaticSamplers, uint32_t flags) {
+    out[0] = 0x44;
+    out[1] = 0x58;
+    out[2] = 0x42;
+    out[3] = 0x43;
+    uint32_t version = 0x00000001;
+    memcpy(out + 4, &version, 4);
```
PR-Reviewer-Ant: This blob is neither a valid DXBC container nor the raw root-signature layout that `D3D12DeviceImpl::CreateRootSignature` parses. Because it starts with `DXBC`, `CreateRootSignature` takes the DXBC path and reads `chunkCount` at byte 28, but this serializer leaves that area zeroed and never emits an `RTS0` chunk table. The result is that any root signature produced by the new `D3D12SerializeRootSignature` parses as zero parameters, so root constants/descriptors silently disappear when a game feeds the serialized blob into `CreateRootSignature`.

include/metalsharp/D3D12Device.h
```diff
@@ -100,11 +101,13 @@ class D3D12FenceImpl final : public ID3D12Fence {
     HRESULT GetCompletedValue(UINT64* pValue) override {
         if (!pValue)
             return E_POINTER;
+        std::lock_guard<std::mutex> lock(m_mutex);
         *pValue = m_completed;
         return S_OK;
```
PR-Reviewer-Ant, line 116: The PR description claims fence sync via wait/wake primitives, but this implementation only adds a mutex around `Signal`/`GetCompletedValue`; `SetEventOnCompletion` still returns `S_OK` without recording or signaling `hEvent`, and `CommandQueue::Wait` remains a no-op. Code that relies on the advertised fence completion behavior can still block forever or race ahead incorrectly.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `fec5d129`

</details>

I found a root-signature serialization incompatibility that makes the new entry point unusable with this repository's own CreateRootSignature parser. There is also a claimed fence-sync change that does not actually implement event completion or waiting.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The serializer writes a custom DXBC-looking header in src/d3d/d3d12/EntryPoint.cpp, while the included CreateRootSignature implementation in include/metalsharp/D3D12Device.h expects either a real DXBC chunk table with an RTS0 chunk or a raw root-signature blob starting with version/param counts. The mismatch is directly visible from the changed code and does not require runtime reproduction.

src/d3d/d3d12/EntryPoint.cpp
```diff
@@ -18,4 +21,111 @@ HRESULT D3D12CreateDevice(void* pAdapter, UINT MinimumFeatureLevel, const GUID&
...
+        samplerData[0] = sampler.Filter;
+        samplerData[1] = sampler.AddressU | (sampler.AddressV << 4) | (sampler.AddressW << 8);
+        float mipBias = sampler.MipLODBias;
+        memcpy(&samplerData[2], &mipBias, 4);
+        samplerData[3] = sampler.MaxAnisotropy | (sampler.ComparisonFunc << 8);
+        float minLod = sampler.MinLOD;
+        memcpy(&samplerData[4], &minLod, 4);
+        float maxLod = sampler.MaxLOD;
+        memcpy(&samplerData[5], &maxLod, 4);
+        samplerData[6] = sampler.ShaderRegister | (sampler.RegisterSpace << 16);
```
PR-Reviewer-Ant, line 172: This returns a raw `malloc` buffer as `ppBlob`, but D3D12 callers expect an `ID3DBlob`-style COM object from `D3D12SerializeRootSignature`, then call methods like `GetBufferPointer()`/`GetBufferSize()` on it. The new tests pass because they treat `blob` as raw bytes and `free()` it, but real Windows code using the normal API contract will interpret the first bytes of this serialized data as a vtable pointer and crash or read garbage. If this project intentionally uses a raw internal root-signature blob, it needs a separate internal helper; the exported D3D12 entry point should return a blob object.

src/d3d/d3d12/EntryPoint.cpp
```diff
@@ -18,4 +21,111 @@ HRESULT D3D12CreateDevice(void* pAdapter, UINT MinimumFeatureLevel, const GUID&
...
+        samplerData[7] = sampler.ShaderVisibility;
+        memcpy(blob + offset, samplerData, 32);
+        offset += 32;
+    }
```
PR-Reviewer-Ant, line 176: `D3D12SerializeVersionedRootSignature` is exported with an extra `UINT Version` parameter. The Windows D3D12 entry point takes the versioned descriptor plus `ppBlob` and `ppErrorBlob`; callers will pass three arguments, while this implementation reads four. On the normal ABI that means the real `ppBlob` value is consumed as `Version`, the real `ppErrorBlob` is treated as `ppBlob`, and the error-blob pointer is whatever happens to be in the next argument register. That makes the new entry point unsafe for the exact compatibility scenario this PR is adding.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `adf551e9`

</details>

I found two D3D12 entry-point compatibility issues in the new root-signature serialization path. Both are likely to break real callers even though the local tests pass because the tests exercise a raw internal blob shape rather than the Windows-facing ABI.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed entry points are fully visible in src/d3d/d3d12/EntryPoint.cpp. The serializer returns a malloc buffer through a blob out-parameter, and the versioned serializer is exported with a four-argument signature that does not match the D3D12 API shape described by the PR.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

src/d3d/d3d12/EntryPoint.cpp
```diff
@@ -18,4 +21,111 @@ HRESULT D3D12CreateDevice(void* pAdapter, UINT MinimumFeatureLevel, const GUID&
...
     return hr;
 }
+
+static void writeRootSignatureHeader(uint8_t* out, uint32_t numParameters, uint32_t numStaticSamplers, uint32_t flags) {
+    uint32_t version = 0x00000001;
+    memcpy(out, &version, 4);
+    memcpy(out + 4, &numParameters, 4);
+    memcpy(out + 8, &numStaticSamplers, 4);
+    memcpy(out + 12, &flags, 4);
+}
```
PR-Reviewer-Ant, line 68: `D3D12SerializeRootSignature` is returning a raw `malloc` buffer through `ppBlob`, but the D3D12 API returns an `ID3DBlob*`. Real callers commonly do `blob->GetBufferPointer()` / `GetBufferSize()` and release the blob as a COM object, so this will turn a successful serialization into an immediate invalid vtable dereference. The test only feeds the raw pointer back into this project's `CreateRootSignature`, so it misses the ABI contract that games actually use.

src/d3d/d3d12/EntryPoint.cpp
PR-Reviewer-Ant, line 176: `D3D12SerializeVersionedRootSignature` has the wrong exported ABI: the Windows function takes `(const D3D12_VERSIONED_ROOT_SIGNATURE_DESC*, ID3DBlob**, ID3DBlob**)`, but this implementation adds a `UINT Version` parameter. A real binary calling the standard three-argument entry point will have its `ppBlob` interpreted as `Version` and its error blob pointer interpreted as `ppBlob`, so the call can corrupt outputs or fail unpredictably before it ever serializes anything.

include/metalsharp/D3D12Device.h
```diff
@@ -174,7 +219,14 @@ class D3D12CommandQueueImpl final : public ID3D12CommandQueue {
             return E_INVALIDARG;
         return pFence->Signal(Value);
     }
-    HRESULT Wait(ID3D12Fence* pFence, UINT64 Value) override { return S_OK; }
+    HRESULT Wait(ID3D12Fence* pFence, UINT64 Value) override {
```
PR-Reviewer-Ant: `ID3D12CommandQueue::Wait` should enqueue a GPU-side wait and return; this implementation blocks the calling CPU thread until the fence value is reached. That changes valid D3D12 ordering into a potential deadlock, for example when a game records `queue->Wait(fence, value)` before submitting later work that will eventually signal that value. This is especially risky because the PR advertises fence sync as a compatibility improvement.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `0716f019`

</details>

I found two D3D12 entry point issues that will break real callers, plus a queue fence wait implementation that can hang applications using normal D3D12 synchronization patterns.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed D3D12 entry points and fence behavior are visible directly in src/d3d/d3d12/EntryPoint.cpp and include/metalsharp/D3D12Device.h. The PR explicitly claims root signature serialization and fence sync support, and these implementations diverge from the D3D12 API contract in ways that are reachable by games.


**Since my last review:**
- Still blocking: 0
- Newly found: 3

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `7f53a432`

</details>

I found two correctness issues in the new D3D12 work: queue waits can break fence ordering, and serialized root signatures are not parsed consistently by the device path. Both are directly in the new implementation and can affect real D3D12 callers.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed diff includes both the new command queue wait execution path and the new root-signature serializer. The ordering bug follows from ExecuteCommandLists returning immediately after detaching the delayed work, while Signal remains synchronous. The root-signature mismatch follows from the serializer writing 20-byte parameter records while the existing parser advances 16 bytes for descriptor/root-descriptor parameters.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

src/d3d/d3d12/D3D12CommandList.mm
```diff
@@ -1098,6 +1101,36 @@ void execute(MetalDevice& metalDev) {
...
+
+        MetalDevice* queuedMetalDevice = &metalDevice;
+        std::thread([queuedMetalDevice, waits = std::move(waits), lists = std::move(lists)]() mutable {
+            for (auto& wait : waits) {
+                wait.first->SetEventOnCompletion(wait.second, nullptr);
+                wait.first->Release();
+            }
+            for (auto* list : lists) {
+                list->__execute(queuedMetalDevice);
+                list->Release();
```
PR-Reviewer-Ant, line 1226: This breaks D3D12 queue ordering when a queued wait is present. `ExecuteCommandLists` moves the work to a detached CPU thread and returns `S_OK` immediately; a following `ID3D12CommandQueue::Signal` on the same queue will therefore complete its fence before these command lists have waited and executed. In D3D12, commands submitted after a queue wait still need to preserve queue order. A common pattern like `queue->Wait(fenceA, v); queue->ExecuteCommandLists(...); queue->Signal(fenceB, v2);` will let `fenceB` signal too early here, which can make the app reuse resources before the delayed command list has actually run.

src/d3d/d3d12/EntryPoint.cpp
```diff
@@ -18,4 +57,113 @@ HRESULT D3D12CreateDevice(void* pAdapter, UINT MinimumFeatureLevel, const GUID&
...
+        if (param.ParameterType == 0) {
+            paramData[2] = param.DescriptorTable.NumDescriptorRanges;
+        } else if (param.ParameterType == 1) {
+            paramData[2] = param.Constants.ShaderRegister;
+            paramData[3] = param.Constants.RegisterSpace;
+            paramData[4] = param.Constants.Num32BitValues;
+        } else {
+            paramData[2] = param.Descriptor.ShaderRegister;
+            paramData[3] = param.Descriptor.RegisterSpace;
+        }
```
PR-Reviewer-Ant, line 144: The serializer writes every root parameter as a 20-byte record, but `CreateRootSignature`'s parser advances only 16 bytes for descriptor tables and root descriptors. That means serialized signatures with descriptor tables, CBVs, SRVs, or UAVs will desynchronize the parser as soon as there is a following parameter/range, even though the PR claims these blobs are compatible with the root-signature path. The new tests only cover a single 32-bit-constants parameter, so this misses the cases D3D12 games commonly use.

include/metalsharp/FormatTranslation.h
```diff
@@ -15,26 +15,117 @@ namespace metalsharp {
...
     DXGI_FORMAT_R8G8B8A8_UNORM = 28,
     DXGI_FORMAT_R8G8B8A8_UNORM_SRGB = 29,
-    DXGI_FORMAT_B8G8R8A8_UNORM = 87,
-    DXGI_FORMAT_B8G8R8A8_UNORM_SRGB = 91,
-    DXGI_FORMAT_R16G16B16A16_FLOAT = 10,
-    DXGI_FORMAT_R32G32B32A32_FLOAT = 2,
+    DXGI_FORMAT_R8G8B8A8_UINT = 30,
+    DXGI_FORMAT_R8G8B8A8_SNORM = 31,
+    DXGI_FORMAT_R8G8B8A8_SINT = 32,
+    DXGI_FORMAT_R16G16B16A16_UNORM_SRGB = 35,
```
PR-Reviewer-Ant: This new enum entry is not a DXGI format and it takes value 35, which is the real `DXGI_FORMAT_R16G16_UNORM`. That means any game/resource using DXGI format 35 will be translated as this invented `R16G16B16A16_UNORM_SRGB` case instead of `R16G16_UNORM`, so the advertised expanded format coverage will mis-create textures for a valid common format. The table needs to match the canonical DXGI numeric values exactly; aliases are fine, but they cannot occupy real format IDs.

src/d3d/d3d12/D3D12CommandList.mm
```diff
@@ -1097,7 +1100,95 @@ void execute(MetalDevice& metalDev) {
...
+    auto state = m_queueState;
+    bool hasOutstandingWork = false;
+    {
+        std::lock_guard<std::mutex> lock(state->mutex);
+        state->stopping = true;
+        hasOutstandingWork = state->busy || !state->work.empty();
+    }
+    state->cond.notify_all();
+    if (state->worker.joinable()) {
+        if (hasOutstandingWork)
```
PR-Reviewer-Ant: Detaching the queue worker while it still has outstanding work leaves queued operations running after `D3D12CommandQueueImpl` destruction. Those operations capture raw pointers/references such as `queuedMetalDevice = &metalDevice` in `ExecuteCommandLists`, but the queue does not own or retain the device, so releasing the queue/device with pending work can turn into a use-after-free. The destructor should drain/join or the queued state must retain the objects it needs until the worker is done.

src/d3d/d3d12/EntryPoint.cpp
```diff
@@ -18,4 +57,120 @@ HRESULT D3D12CreateDevice(void* pAdapter, UINT MinimumFeatureLevel, const GUID&
...
+}
+
+HRESULT D3D12SerializeRootSignature(const D3D12_ROOT_SIGNATURE_DESC* pRootSignatureDesc, UINT Version, void** ppBlob,
+                                    void** ppErrorBlob) {
+    if (!pRootSignatureDesc || !ppBlob)
+        return E_INVALIDARG;
+    if (Version > 0x1)
+        return E_INVALIDARG;
+
+    uint32_t numParams = pRootSignatureDesc->NumParameters;
```
PR-Reviewer-Ant, line 109: The PR description says this generates DXBC-compatible root signature blobs, but the implementation writes a private raw layout starting with version/parameter counts rather than a DXBC container with an `RST0` chunk. That works only because `CreateRootSignature` also understands this private layout; it is not the compatibility surface the PR claims. If the intent is just internal round-tripping, the description should be narrowed, but if games/tools expect a normal D3D blob this needs a real DXBC container.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `6c1558ed`

</details>

I found two correctness issues that should be fixed before this lands: the expanded DXGI table introduces at least one non-DXGI enum value that collides with a real format, and the async command queue can detach work that still depends on the queue/device lifetime.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both issues are visible directly in the changed code: the new format enum assigns value 35 to a nonstandard R16G16B16A16 sRGB entry, while DXGI value 35 is R16G16_UNORM; the command queue destructor explicitly detaches the worker when outstanding work exists while queued lambdas retain only raw pointers/references.


**Since my last review:**
- Still blocking: 0
- Newly found: 3

include/metalsharp/FormatTranslation.h
```diff
@@ -15,26 +15,116 @@ namespace metalsharp {
...
+    DXGI_FORMAT_AYUV = 100,
+    DXGI_FORMAT_NV12 = 103,
+    DXGI_FORMAT_P010 = 104,
+    DXGI_FORMAT_YUY2 = 107,
+    DXGI_FORMAT_NV11 = 108,
+    DXGI_FORMAT_R10G10B10A2_TYPELESS_COMPAT = 118,
 };
 
+constexpr DXGITranslation DXGI_FORMAT_D32_FLOAT = DXGITranslation::DXGI_FORMAT_R32_TYPELESS;
+
```
PR-Reviewer-Ant, line 126: This aliases `DXGI_FORMAT_D32_FLOAT` to `R32_TYPELESS` value 39, but the public D3D headers still define D32 as value 40. Any D3D12 resource created with `pDesc->Format == 40` gets cast to `DXGITranslation(40)`, which is not listed in this enum, so the new `dxgiFormatIsDepth()` check fails and `dxgiFormatToMetal()` returns `MTLPixelFormatInvalid`. That breaks the PR's claimed D32/depth coverage for normal callers using the D3D12 constants.

src/d3d/d3d12/D3D12CommandList.mm
```diff
@@ -1097,7 +1100,90 @@ void execute(MetalDevice& metalDev) {
     return S_OK;
 }
 
+D3D12CommandQueueImpl::~D3D12CommandQueueImpl() {
+    auto state = m_queueState;
+    {
+        std::lock_guard<std::mutex> lock(state->mutex);
+        state->stopping = true;
+    }
```
PR-Reviewer-Ant, line 1109: `ID3D12CommandQueue::Wait` now enqueues a worker operation that blocks inside `pFence->SetEventOnCompletion(Value, nullptr)`. If the queue is released while that wait is pending, the destructor sets `stopping = true` and then joins the worker, but the worker cannot observe `stopping` until the fence is signaled. That can hang shutdown/destruction indefinitely for an unsatisfied external fence wait.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `dd6abc5c`

</details>

I found two correctness issues in the new D3D12 path: D32 depth resources no longer translate correctly from the public DXGI value, and the queued fence wait can deadlock queue teardown.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The D32 mismatch is visible across include/d3d/D3D12.h, include/metalsharp/FormatTranslation.h, and the changed CreateCommittedResource format path. The queue deadlock follows directly from Wait enqueueing a blocking SetEventOnCompletion call and the destructor joining the worker without a way to interrupt that active wait.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

src/d3d/d3d12/EntryPoint.cpp
```diff
@@ -18,4 +57,142 @@ HRESULT D3D12CreateDevice(void* pAdapter, UINT MinimumFeatureLevel, const GUID&
...
+    constexpr uint32_t headerSize = 32;
+    constexpr uint32_t chunkCount = 1;
+    constexpr uint32_t chunkTableSize = chunkCount * 4;
+    constexpr uint32_t chunkOffset = headerSize + chunkTableSize;
+    uint32_t chunkSize = static_cast<uint32_t>(rootSignature.size());
+    uint32_t totalSize = chunkOffset + 8 + chunkSize;
+
+    std::vector<uint8_t> container(totalSize);
+    memcpy(container.data(), &dxbcMagic, 4);
+    memcpy(container.data() + 24, &totalSize, 4);
```
PR-Reviewer-Ant, line 86: This walks `pRootSignatureDesc->pParameters[i]` whenever `NumParameters` is nonzero, but it never checks that `pParameters` is non-null. The same pattern appears later for descriptor-table ranges and static samplers. A game or middleware passing an invalid root signature descriptor should get `E_INVALIDARG`/an error blob, not a host-side crash in the shim. Please validate `NumParameters > 0 => pParameters != nullptr`, `NumStaticSamplers > 0 => pStaticSamplers != nullptr`, and descriptor-table `NumDescriptorRanges > 0 => pDescriptorRanges != nullptr` before dereferencing.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `4d17b4b9`

</details>

I found one correctness issue in the new root signature serializer: malformed descriptors can crash the process instead of returning an error. The rest of the diff largely matches the stated D3D12/DXGI expansion, with the caveat that the runtime HDR/game smoke tests are still unchecked.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The null dereference is directly visible in src/d3d/d3d12/EntryPoint.cpp: the serializer indexes pParameters and pDescriptorRanges based only on the count fields. I did not have a runtime trace from a real game or shader compiler path to validate the broader DXBC compatibility claims.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

src/d3d/d3d12/D3D12CommandList.mm
```diff
@@ -1097,7 +1100,114 @@ void execute(MetalDevice& metalDev) {
...
+    state->cond.notify_one();
+    return S_OK;
+}
+
+HRESULT D3D12CommandQueueImpl::Signal(ID3D12Fence* pFence, UINT64 Value) {
+    if (!pFence)
+        return E_INVALIDARG;
+
+    if (!hasQueuedWork())
+        return pFence->Signal(Value);
```
PR-Reviewer-Ant, line 1160: This new worker-thread `Release()` makes object lifetime racy. The D3D12 COM objects still use plain `ULONG refCount` with unsynchronized `++refCount` / `--refCount`, so a game thread releasing the same fence or command list while the queue worker releases its retained reference can lose updates or delete the object too early. Since this PR moves fence signaling/waits and queued command-list execution onto a background thread, the affected refcounts need to become thread-safe before this is safe.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `71adf209`

</details>

I found a concurrency issue in the new asynchronous D3D12 queue path: it moves COM Release calls onto a worker thread while the underlying refcounts are still plain non-atomic integers. That can corrupt object lifetime under normal multi-threaded D3D12 usage.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new queue worker explicitly releases fences and command lists from a background thread in src/d3d/d3d12/D3D12CommandList.mm, while the COM implementations in include/metalsharp/D3D12Device.h still use plain ULONG refCount fields with ++/--. This is directly introduced by the async queue changes.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

src/d3d/d3d12/EntryPoint.cpp
```diff
@@ -18,4 +58,150 @@ HRESULT D3D12CreateDevice(void* pAdapter, UINT MinimumFeatureLevel, const GUID&
...
+
+HRESULT D3D12GetDebugInterface(const GUID& riid, void** ppvDebug) {
+    if (!ppvDebug)
+        return E_POINTER;
+    *ppvDebug = nullptr;
+    return E_NOINTERFACE;
+}
+
+HRESULT D3D12EnableExperimentalFeatures(UINT NumFeatures, const GUID* pFeatureGUIDs, void* pConfigurationStructs,
+                                        UINT* pConfigurationStructSizes) {
```
PR-Reviewer-Ant, line 204: Returning `S_OK` unconditionally from `D3D12EnableExperimentalFeatures` is a misleading contract. A caller can pass non-zero feature GUIDs and then assume those experimental features are enabled, but this implementation neither validates `pFeatureGUIDs` nor enables anything. This should reject unsupported/non-empty requests with an error such as `E_NOINTERFACE`/`E_INVALIDARG`, while only treating the no-feature case as a harmless no-op.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `94b862b5`

</details>

I found one correctness issue in the D3D12 entry-point stubs: experimental features now report success even when the inputs are invalid or unsupported, which can make callers take paths MetalSharp does not actually enable.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The changed entry point is explicit in src/d3d/d3d12/EntryPoint.cpp and is part of the PR's stated new API surface. I did not run a game/runtime repro, so the exact downstream failure mode depends on which title probes experimental D3D12 features.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `98fc168d`

</details>

I found one correctness issue in the advertised format coverage: the swapchain path maps RGBA formats to a BGRA Metal layer, which can silently swap channels for callers that requested RGBA.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The PR description specifically claims format-aware swapchain pixel format selection, and the new switch in src/dxgi/DXGISwapChain.mm is directly in that path. I did not have a runtime HDR/swapchain repro, so confidence is limited to the static DXGI-to-Metal mapping evidence.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

src/dxgi/DXGISwapChain.mm
```diff
@@ -18,6 +19,29 @@
 #include <Metal/Metal.h>
 #include <QuartzCore/QuartzCore.h>
 
+static MTLPixelFormat dxgiFormatToSwapchainMetal(DXGI_FORMAT format) {
+    switch (format) {
+    case DXGI_FORMAT_R8G8B8A8_UNORM:
+    case DXGI_FORMAT_R8G8B8A8_UNORM_SRGB:
```
PR-Reviewer-Ant, line 26: This maps `DXGI_FORMAT_R8G8B8A8_UNORM` and `_SRGB` swapchains to `MTLPixelFormatBGRA8Unorm`. That is not just a storage detail: a caller that requested RGBA will get a BGRA drawable, so red/blue channels can be swapped when presenting. The B8G8R8A8 cases below are the ones that should use BGRA; RGBA swapchains need an RGBA-compatible Metal format or a conversion path before presenting.

src/d3d/d3d12/D3D12CommandList.mm
```diff
@@ -1097,7 +1101,114 @@ void execute(MetalDevice& metalDev) {
...
+
+    pFence->AddRef();
+    return enqueueQueueWork([pFence, Value] {
+        pFence->Signal(Value);
+        pFence->Release();
+    });
+}
+
+HRESULT D3D12CommandQueueImpl::Wait(ID3D12Fence* pFence, UINT64 Value) {
```
PR-Reviewer-Ant, line 1171: This `Wait` implementation can deadlock the command queue. A queued wait runs on the single queue worker and spins until the fence reaches `Value`; while it is spinning, no later queued work can run. If the waited fence is satisfied by a later `Signal` or command-list completion on this same queue, that signal is stuck behind the wait forever. D3D12 queue waits need to order GPU work without blocking the queue's own worker thread, or at least detect same-queue waits and handle them without preventing the satisfying work from executing.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `e18d16cd`

</details>

I found one correctness issue in the new queue fence path. The implementation can deadlock a command queue when a wait is satisfied by later work on the same queue.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The queue worker is introduced in this diff, and the blocking wait loop is directly visible in D3D12CommandList.mm. The behavior follows from the worker executing queued operations serially.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d0e1a752`

</details>

I did not find a concrete correctness, security, data integrity, or breaking-change issue in the changed paths. The implementation broadly matches the PR description, with the remaining risk concentrated in runtime-only D3D12/HDR behavior that the test plan already marks as not smoke-tested.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the relevant D3D12 entry point, fence/queue, format translation, DXGI swapchain, and focused tests, and I checked the implementation against those claims. I could not validate the untested runtime cases called out in the PR description, especially real game launch, Steam CEF swapchain negotiation, and HDR display output.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #104, opened 2026-05-21
[codex] Bind DXMT winemetal through Wine runtime

aaf2tbz:
## Summary
- Bind the DXMT `winemetal` PE/Unix pair through the Wine runtime instead of copying `winemetal.dll` beside each game executable.
- Require and preflight `dxmt/winemetal.dll` plus `dxmt/winemetal.so` for M10/M11/M12 recipes, and log the prefix/runtime binding paths during launches.
- Force DXMT-owned DirectX DLLs to native-only load order and provide `WINEDLLPATH` for DXMT routes so Wine cannot silently blend DXMT `dxgi.dll` with builtin `d3d11.dll`/WineD3D.
- Carry Steam identity (`SteamAppId` and `SteamGameId`) into direct Wine game launches so Steamworks titles do not bounce back through an unprepared Steam confirmation handoff.
- Harden the shared M9/M10/M11/M12 launch contract: create route/appid shader and pipeline cache dirs before launch, export the same DXMT cache env, and force M11/M12 to the 70% render-scale MetalFX path via `d3d11.metalSpatialUpscaleFactor=1.43`.
- Keep Unity D3D forcing route-aware: M10/M11 force D3D11, while explicit M12 Unity launches auto-add `-force-d3d12` unless the user already supplied a renderer flag.
- Record the DXMT patch set needed for Unity/DXMT startup: Wine private DXGI D3D10 handoff, no-op `ID3D12InfoQueue`, typed-UAV feature parity, default shader-model cap handling, and per-launch D3D12 trace-file routing.
- Add installer evidence hardening for EA/Origin, Ubisoft, and EAC: installer launch logs now include classifier/profile/pipeline/hint metadata, installer-focused Wine debug env, direct MSI verbose logs, and `/launcher/evidence` now reports EAC EOS install/product-id/module-load evidence without adding bypass behavior.

## Why
The current M11/M12 work exposed several launch failures that looked similar from the app but came from different layers:

1. Subnautica: Below Zero on M11 could map DXMT DLLs but then blend in builtin WineD3D, leaving the process alive without a useful graphics stack.
2. Unity/DXMT routes needed Wine's private `DXGID3D10CreateDevice` / `DXGID3D10RegisterLayers` exports from DXMT `dxgi.dll` for D3D11 compatibility startup.
3. Sons Of The Forest on a forced M12 probe reached DXMT D3D12, but Unity queried `ID3D12InfoQueue`; without the no-op surface it collapsed into the generic DirectX 11 initialization error even after D3D12 device creation.
4. After the InfoQueue patch, Unity's forced M12 path reached HDRP compute-kernel selection. The local trace showed `TypedUAVLoadAdditionalFormats = FALSE` even though per-format support reports typed UAV load/store support, and default shader-model probes could observe an accidental zero cap.
5. Installer and launcher failures need durable bottle evidence. EA/Origin, Ubisoft, and EAC now produce enough launch/evidence metadata to tell MSI/service/runtime failures apart from WebView, CEF, and protected-launch module failures.

This PR makes the launcher path more authoritative, records the runtime patches needed for future bundle rebuilds, keeps cache/upscale behavior consistent across the new Wine runtime, and improves installer evidence without changing anti-cheat compatibility boundaries.

## Current Runtime Result
- Subnautica: Below Zero M11 no longer loads `wined3d.dll` after native-only DXMT overrides and advances into Unity/asset loading with `winemetal.so`.
- Sons Of The Forest M12 diagnostic run now gets past the old `d3d12: no D3D12 installed` wall.
- With the InfoQueue patch installed locally, trace output shows `D3D12CreateDevice`, `D3D11On12CreateDevice`, and repeated M12 swapchain presents through a CAMetalLayer at 1920 x 1080.
- With typed-UAV and shader-model patches installed locally, trace output shows `TypedUAV=1` and `SHADER_MODEL: HighestSM=96`.
- M9/M10/M11/M12 now share explicit MetalSharp/DXMT shader and pipeline cache paths under `~/.metalsharp/{shader-cache,pipeline-cache}/<route>/<appid>/`.
- M11/M12 launches now repair/export the DXMT config needed for MetalFX swapchain upscaling from a 70% internal render target.
- EA/Origin, Ubisoft, and EAC installer bottles now preserve richer launch logs and evidence reports for the next installer proof run.

## Validation
- `cargo fmt --manifest-path app/src-rust/Cargo.toml --all -- --check`
- `cargo test --manifest-path app/src-rust/Cargo.toml mtsp::launcher:: -- --nocapture`
- `cargo test --manifest-path app/src-rust/Cargo.toml mtsp::engine:: -- --nocapture`
- `cargo test --manifest-path app/src-rust/Cargo.toml mtsp::recipe:: -- --nocapture`
- `cargo test --manifest-path app/src-rust/Cargo.toml launcher_evidence:: -- --nocapture`
- `cargo test --manifest-path app/src-rust/Cargo.toml bottles::tests:: -- --nocapture`
- `cargo test --manifest-path app/src-rust/Cargo.toml sharp_library:: -- --nocapture`
- `git diff --check`
- `git -C /Volumes/AverySSD/metalsharp/dxmt-src apply --reverse --check tools/wine/patches/dxmt-d3d12-infoqueue-compat.patch`
- `git -C /Volumes/AverySSD/metalsharp/dxmt-src apply --reverse --check tools/wine/patches/dxmt-d3d12-typed-uav-feature.patch`
- `git -C /Volumes/AverySSD/metalsharp/dxmt-src apply --reverse --check tools/wine/patches/dxmt-d3d12-shader-model-default.patch`
- `tools/wine/apply-dxmt-patches.sh --check /Volumes/AverySSD/metalsharp/dxmt-src`

## Follow-up
The D3D12 blocker has moved from loader/device initialization into real Unity/HDRP runtime behavior. The next useful investigation is render-target sizing, compute dispatch dimensions, and resource/barrier behavior after the Unity route enters the render loop. For launcher installers, the next proof runs should refresh `/launcher/evidence` after EA/Origin, Ubisoft, and EAC installer attempts.


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm taking a look at the changes in `[codex] Bind DXMT winemetal through Wine runtime` (commit `b5e4e286`) and will follow up shortly.

app/src-rust/src/launcher_evidence.rs
```diff
@@ -302,6 +355,24 @@ fn summarize_launcher(
...
+        for asset in &manifest.runtime_assets {
+            let lower = asset.source_path.to_ascii_lowercase();
+            if lower.contains("easyanticheat") {
+                facts.eac_setup_detected = true;
+            }
+            if lower.contains("easyanticheat_eos") || lower.contains("eos") {
+                facts.eac_eos_setup_detected = true;
+            }
+            if lower.ends_with(".so") {
+                facts.eac_proton_asset_present = true;
```
PR-Reviewer-Ant, line 368: This can report `eacProtonAssetPresent=true` for unrelated Unix assets. The surrounding loop only checks that the bottle family is `easyanticheat`, then treats any runtime asset whose `source_path` ends in `.so` as EAC Proton evidence. A bottle that has an unrelated Unix bridge or helper `.so` would be summarized as `eac_installed_with_proton_assets`, which weakens the new `/launcher/evidence` signal the PR is adding. This should probably require an EAC-specific filename/path, e.g. `easyanticheat*.so` or the expected vendor module names.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `b5e4e286`

</details>

I found one evidence-quality bug in the new EAC reporting path. The launcher/runtime changes otherwise line up with the PR description from the diff I reviewed.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The EAC false-positive is directly visible in launcher_evidence.rs: the family-specific loop marks any `.so` runtime asset as EAC Proton evidence. I did not have a live bottle/runtime artifact to prove the observed endpoint output, so I am keeping this as a non-blocking comment.

app/src-rust/src/mtsp/launcher.rs
```diff
@@ -655,19 +1015,81 @@ fn build_dyld(ms_root: &PathBuf, paths: &[&str]) -> String {
...
+        return Ok(None);
+    }
+
+    let config_path = ms_root.join("etc").join("dxmt.conf");
+    if let Some(parent) = config_path.parent() {
+        std::fs::create_dir_all(parent)?;
+    }
+    let current = std::fs::read_to_string(&config_path).unwrap_or_default();
+    if !dxmt_config_matches_contract(&current) {
+        std::fs::write(&config_path, DXMT_CONFIG_CONTENT)?;
```
PR-Reviewer-Ant: `ensure_dxmt_config` replaces the entire shared `~/.metalsharp/runtime/wine/etc/dxmt.conf` whenever the current file is missing one of the three contract lines. That preserves the new MetalFX defaults, but it also silently deletes any existing DXMT/runtime settings the user or bundle had in that file. Since this runs from normal DXMT launch/env preparation, a single launch can corrupt the runtime config for all M9/M10/M11/M12 routes. Please merge or upsert just the required keys while preserving unrelated lines/comments, or move these overrides entirely into the per-launch `DXMT_CONFIG` env that this PR already exports.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `a514813f`

</details>

I found one launch-contract/data-preservation issue: the new DXMT config repair path rewrites the whole shared runtime config when any required key is missing. That can silently discard existing DXMT settings while preparing or launching unrelated DXMT routes.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The config overwrite is explicit in app/src-rust/src/mtsp/launcher.rs and is reached by DXMT launch/env preparation paths. I could not run the repository tests because the checkout was not present in the sandbox, but the behavior follows directly from the supplied diff.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/mtsp/recipe.rs
```diff
@@ -201,6 +210,65 @@ pub fn build_custom_launch_recipe(
     })
 }
 
+fn launch_args_for_recipe<'a>(
+    node: &PipelineNode,
+    game_dir: Option<&Path>,
+    _configured_pipeline: Option<PipelineId>,
+    base_args: impl IntoIterator<Item = &'a str>,
```
PR-Reviewer-Ant, line 217: This ignores `_configured_pipeline` and always adds `-force-d3d12` for any Unity launch that reaches the M12 node. That contradicts the PR's stated route-aware behavior and the updated M12 docs, which say M12 probes of M11-mapped Unity titles should keep user launch args authoritative instead of injecting D3D12. As written, a diagnostic/explicit M12 launch for a title still configured as M11 will get the forced D3D12 flag automatically, reintroducing the exact Unity/HDRP path the PR says it is avoiding. Use the configured route to gate this, or drop the claim/test that M11-mapped titles should not be auto-forced.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f9aa3bd6`

</details>

I found one correctness issue: the Unity launch-arg logic does the opposite of the PR/docs claim for M12 probes of M11-mapped games.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description and docs say M12 Unity renderer forcing should remain route-aware, but app/src-rust/src/mtsp/recipe.rs ignores the configured pipeline and unconditionally injects -force-d3d12 whenever the selected node is M12.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/mtsp/launcher.rs
```diff
@@ -497,7 +738,16 @@ fn launch_wine_bare_with_context(
         crate::platform::runtime_library_env(&ms_root).map(|(key, _)| key).unwrap_or("LD_LIBRARY_PATH");
```
PR-Reviewer-Ant, line 738: This path still does not carry `SteamAppId` / `SteamGameId`. `launch_steam_bottle_with_pipeline` routes `PipelineId::M32 | PipelineId::WineBare` through `launch_wine_bare_with_context`, but `apply_steam_identity_env` is only called in the DXMT path. That means Steamworks titles launched through the plain Wine/32-bit fallback can still bounce through the unprepared Steam handoff, despite the PR description saying direct Wine game launches carry Steam identity.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `610411d2`

</details>

I found a couple of correctness problems that look worth fixing before this lands: the Steam identity handoff is not actually applied to all direct Wine launches, and the DXGI/D3D12 interface expansion appears to leave existing concrete implementations abstract or ABI-mismatched.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The launcher issue is directly visible in mtsp/launcher.rs: Steam bottle M32/WineBare routes call launch_wine_bare_with_context, but SteamAppId/SteamGameId are only applied in the DXMT launch path. The C++ interface issue is based on the diff adding new pure virtual methods in public COM-style interfaces without corresponding implementation changes in the PR diff; I could not run a C++ build in this read-only review environment.


**Since my last review:**
- Still blocking: 0
- Newly found: 3

include/dxgi/DXGI.h
```diff
@@ -35,11 +41,66 @@ class MIDL_INTERFACE("2411e7e1-12ac-4ccf-bd14-9798e8534dc0") IDXGIFactory1 : pub
     STDMETHOD_(INT, IsCurrent)() PURE;
```
PR-Reviewer-Ant, line 41: Adding the full `IDXGIFactory2/3/4` pure virtual surface here requires the concrete DXGI factory implementation to provide every new slot. I do not see corresponding changes to `DXGIFactory`/`DXGIFactory.cpp` in this PR, so any class currently deriving from `IDXGIFactory4` becomes abstract or, worse, has a COM vtable layout that no longer matches what callers expect. Please add stub implementations at least for the new methods before changing the interface.

include/d3d/D3D12.h
```diff
@@ -880,6 +959,8 @@ class ID3D12Device : public ID3D12Object {
     STDMETHOD(CreateCommittedResource)(const D3D12_HEAP_PROPERTIES* pHeapProperties, UINT HeapFlags,
```
PR-Reviewer-Ant, line 959: This inserts new pure virtual methods into `ID3D12Device`, but the diff does not show matching overrides for the concrete D3D12 device class. Since this is a COM-style interface, adding slots without updating the implementation is a build break if the class derives from this header, and an ABI hazard if anything is consumed through the old layout. Please wire `GetResourceAllocationInfo`, `CopyDescriptors`, and `CopyDescriptorsSimple` through the concrete device implementation or keep them out of the public interface for this PR.

include/dxgi/DXGI.h
PR-Reviewer-Ant, line 41: This expands `IDXGIFactory2` from an empty derived interface into a COM interface with many new pure virtual methods. Any concrete factory class that currently implements `IDXGIFactory4` now also has to implement `IsWindowedStereoEnabled`, `CreateSwapChainForHwnd`, `CreateSwapChainForComposition`, the registration methods, and the new `IDXGIFactory3/4` methods. I don't see corresponding implementation changes in the diff, so this is likely to either fail to compile because the factory class becomes abstract, or worse, shift the expected vtable surface without working method bodies. Please add the implementations/stubs alongside the header expansion and cover factory creation with a build/test that instantiates the concrete DXGI factory.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e80ee5bd`

</details>

I found one likely build/ABI break in the C++ COM surface expansion. The Rust launcher changes broadly match the PR description, but the DXGI header change needs implementation coverage before this should land.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff explicitly turns previously empty IDXGIFactory2/3/4 interfaces into interfaces with new pure virtual methods in include/dxgi/DXGI.h. The supplied diff does not show matching DXGIFactory implementation changes, so the risk is directly visible from the changed header, though the prompt truncates some C++ context.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/mtsp/launcher.rs
```diff
@@ -655,19 +1021,122 @@ fn build_dyld(ms_root: &PathBuf, paths: &[&str]) -> String {
...
+    if let Err(err) = std::fs::create_dir_all(path) {
+        eprintln!("metalsharp: failed to create {} cache dir {}: {}", label, path.display(), err);
+    }
+}
+
+fn ensure_dxmt_config(node: &PipelineNode, ms_root: &Path) -> Result<Option<PathBuf>, Box<dyn std::error::Error>> {
+    if node.backend != "dxmt" {
+        return Ok(None);
+    }
+
```
PR-Reviewer-Ant, line 1071: This applies the new `d3d11.metalSpatialUpscaleFactor`, `preferredMaxFrameRate`, and `maxFeatureLevel` config to every DXMT backend route. That includes M9 and M10, because the only guard is `node.backend == "dxmt"`. The PR description says the forced 70% MetalFX path is for M11/M12, while M9/M10 are only supposed to share cache/runtime setup. Since this writes the shared `~/.metalsharp/runtime/wine/etc/dxmt.conf` and `apply_dxmt_config_env` also exports the same `DXMT_CONFIG` for all DXMT routes, a D3D9/D3D10 launch can silently mutate and inherit a D3D11/M11-style render-scale contract. Please scope this to the intended pipelines, or update the contract/tests if M9/M10 are deliberately meant to receive these D3D11 settings too.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `5e9a23c4`

</details>

I found one launch-contract mismatch: the new DXMT config/upscale contract is applied to every DXMT pipeline, not just the M11/M12 routes described by the PR.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The launcher changes gate DXMT config creation and DXMT_CONFIG export only on node.backend == "dxmt", which covers M9/M10/M11/M12. The PR description and docs repeatedly say the forced 70% MetalFX path is for M11/M12; I could verify the changed diff, but the workspace only includes the review prompt rather than the full repository checkout.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

include/d3d/D3D12.h
```diff
@@ -880,6 +959,8 @@ class ID3D12Device : public ID3D12Object {
...
                                        const D3D12_RESOURCE_DESC* pDesc, UINT InitialResourceState,
```
PR-Reviewer-Ant, line 960: This inserts `GetResourceAllocationInfo` into `ID3D12Device` before `CreateDescriptorHeap`, which changes the vtable slot of every method after `CreateCommittedResource`. D3D12 is a COM ABI surface, so callers compiled against the real Windows `ID3D12Device` layout will dispatch to the wrong methods once they reach descriptor heaps, root signatures, command signatures, or descriptor copies. If this interface is meant to be ABI-compatible, new methods need to be added in the exact Windows SDK order, with the existing later methods kept in their real slots.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `230c2ec5`

</details>

I found one native ABI issue in the D3D12 surface that should be fixed before this lands. The launcher/runtime changes broadly match the PR description, but the COM interface edit is risky in a way that will not be caught by ordinary compile-only coverage.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed header shows new ID3D12Device virtual methods inserted into the interface body, which is enough to identify a COM vtable compatibility problem. I did not have a runtime trace from a Windows title calling through this interface, so the confidence is medium rather than high.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/mtsp/recipe.rs
```diff
@@ -747,9 +817,28 @@ fn runtime_assets_for_node(node: &PipelineNode, ms_root: &Path) -> Vec<RuntimeAs
...
     if node.backend == "dxmt" {
+        let pe = ms_root.join("lib").join("dxmt").join("x86_64-windows").join("winemetal.dll");
+        assets.push(RuntimeAsset { name: "dxmt/winemetal.dll".into(), present: pe.exists(), path: pe, required: true });
+        let unix = ms_root.join("lib").join("dxmt").join("x86_64-unix").join("winemetal.so");
+        assets.push(RuntimeAsset {
+            name: "dxmt/winemetal.so".into(),
+            present: unix.exists(),
+            path: unix,
+            required: true,
+        });
```
PR-Reviewer-Ant, line 839: This makes `dxmt.conf` required for every DXMT backend route, including M9 and M10. But the launcher contract added in this PR is explicitly M11/M12-only: `dxmt_config_applies()` returns true only for M11/M12, and the new tests assert M9/M10 do not export `DXMT_CONFIG_FILE` or `DXMT_CONFIG`. As written, runtime asset preflight/doctor can now report M9/M10 as missing a required `dxmt.conf` even though those routes will never create or use it. Please either make the asset required only when `dxmt_config_applies(node)` is true, or keep it optional for M9/M10.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `51e1417b`

</details>

I found one runtime-contract regression: M9/M10 now require `dxmt.conf` in recipe preflight even though the launcher intentionally does not create or export the DXMT config contract for those routes.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The diff marks `dxmt.conf` as a required asset for every `node.backend == "dxmt"` recipe, while the same PR adds tests and launcher logic showing `DXMT_CONFIG_FILE` is M11/M12-only. That makes the regression directly visible from the changed recipe and launcher code.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

include/dxgi/DXGI.h
```diff
@@ -35,11 +41,66 @@ class MIDL_INTERFACE("2411e7e1-12ac-4ccf-bd14-9798e8534dc0") IDXGIFactory1 : pub
...
 };
```
PR-Reviewer-Ant, line 43: This turns IDXGIFactory2/3/4 from empty derived interfaces into interfaces with new pure virtual methods, but the PR diff does not include matching DXGIFactory implementation changes. Any concrete factory class that already advertises IDXGIFactory4 now has to implement CreateSwapChainForHwnd/CoreWindow/Composition, GetCreationFlags, EnumAdapterByLuid, EnumWarpAdapter, etc.; otherwise the native target will fail to compile once this header is used. Since this PR also changes CMake to link dxgi against d3d12, please either add the required implementations/stubs or keep these methods out of the abstract interface until the implementation is in place.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `c95024d9`

</details>

I found a likely C++ build break in the DXGI interface expansion: the PR adds new pure virtual IDXGIFactory2/3/4 methods but does not show corresponding DXGIFactory implementation changes. The Rust launcher changes otherwise line up with the PR description at the diff level.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The diff explicitly changes include/dxgi/DXGI.h to make previously empty IDXGIFactory2/3/4 interfaces contain pure virtual methods, while no src/dxgi/DXGIFactory implementation changes are included in the diff and the PR validation does not list a C++ build. I could not run the native build in this read-only review environment.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `1683b79b`

</details>

I did not find a concrete correctness, security, or data-integrity issue worth flagging from the provided diff. The implementation appears aligned with the PR description around DXMT winemetal runtime binding, native-only DirectX overrides, Steam identity propagation, cache/config setup, Unity route-aware args, and launcher evidence expansion.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The Rust launch, recipe, evidence, and docs changes include targeted tests for the new contracts, and the changed lines directly support the PR claims. I could not run the runtime launch paths or inspect the truncated native D3D12 implementation beyond the provided diff, so confidence is medium rather than high.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

app/src-rust/src/sharp_library.rs
```diff
@@ -566,9 +567,10 @@ fn start_wine_installer_in_bottle(
     let launch_id = installer_launch_id(src, pipeline);
```
PR-Reviewer-Ant, line 567: This writes the new classifier/profile/pipeline metadata before the MSI branch, but `launch_msi_installer` then owns the same `log_path` and writes its own MSI header/output. If that function opens the file with truncating semantics, the metadata this PR promises for installer launch logs is lost specifically for MSI installers, which are one of the paths called out in the description. Please make the MSI path append to the existing log or move the metadata write inside the MSI log creation flow so MSI evidence keeps the same fields as EXE launches.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `79cc6fb9`

</details>

I found one issue in the installer evidence path: the new metadata is written before the MSI-specific launcher takes over, but the MSI log path can still be reopened in a way that loses that metadata. That undercuts one of the PR's stated validation goals for MSI installer evidence.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The PR adds write_installer_launch_metadata before branching into launch_msi_installer in app/src-rust/src/sharp_library.rs, and the MSI path separately initializes the same launch log. I could verify the flow from the provided diff, but the workspace only includes the packaged review prompt rather than a complete checkout, so I could not run the Rust tests or inspect any unshown helper changes.


**Since my last review:**
- Fixed or likely addressed: 0
- Still blocking: 0
- Newly found: 1

PR #115, opened 2026-05-23
Beta 7 foundational runtime work

aaf2tbz:
## Beta 7 Foundational Runtime Work

Comprehensive runtime improvements — DLL routing, GPTK pipeline, redistributables, GPU stubs, font registry, DX verification, per-game recipes, auto-install.

### Completed Phases (16/17)

| Phase | Description |
|-------|-------------|
| **1A** | WINEDLLPATH routing for DXMT M9-M12 |
| **1B** | M13 GPTK launch routing |
| **1B.5** | Apple GPTK 3.0 bundled (D3DMetal framework) |
| **2A** | VC++ 2015-2022 expanded detection |
| **2B** | VC++ 2013 redistributable |
| **2C** | DirectX Jun 2010 full verification — 39 DLL inventory |
| **2D** | GPU vendor stubs (nvapi64, nvngx, atidxx64) |
| **3A** | Font substitution registry — 21 HKLM + 10 HKCU |
| **3B** | Declarative bottle templates |
| **3C** | Drive mappings (y: → user home) |
| **4A/B** | Anti-cheat clean + offline mode |
| **4C** | EOS SDK env-based offline — research complete (no env var exists) |
| **5A** | Post-wineboot config seeding |
| **5B** | Per-game TOML recipe expansion — 12 titles with dependencies |
| **5C** | CrossTie-inspired install recipes — `POST /steam/install-recipe-deps` |
| **Bundle** | DXMT + GPTK tarballs uploaded to bundles release |

### Remaining (1/17 — Deferred)

| Phase | Description |
|-------|-------------|
| 1C | mscompatdb.so v2 — Wine ntdll interception layer (long-term) |

### New Endpoints
- `POST /bottles/verify-directx` — inventory 39 DX DLLs per prefix
- `POST /steam/install-recipe-deps` — auto-install missing recipe dependencies
- `POST /bottles/apply-font-subs` — font substitution registry
- `POST /bottles/seed-post-wineboot` — prefix config seeding

### Verification
- `cargo fmt --all -- --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `cargo test` — 158 tests pass ✅
- `npm run build` ✅
- Bundle tarballs uploaded to `bundles` release ✅


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm taking a look at the changes in `Beta 7 foundational runtime work` (commit `fec634ad`) and will follow up shortly.

app/src-rust/src/installer.rs
```diff
@@ -504,6 +506,65 @@ fn install_dxmt_runtime(home: &PathBuf) -> Result<bool, String> {
...
+                } else {
+                    let _ = fs::copy(entry.path(), &dst);
+                }
+            }
+        }
+
+        let _ = fs::remove_dir_all(&tmp);
+    }
+
+    if gptk_dir.join("x86_64-windows").join("d3d11.dll").exists() && framework.exists() {
```
PR-Reviewer-Ant: This returns `Ok(false)` when the GPTK bundle is absent or incomplete, so `/setup/install-all` will mark "GPTK D3DMetal Runtime" as ready even though M13 cannot work. That makes the new setup path silently skip the PR's advertised bundled GPTK runtime. This should fail the step, or at least report an explicit missing/skipped state that the setup UI and runtime doctor can surface.

app/src-rust/src/main.rs
```diff
@@ -698,6 +698,14 @@ fn route(req: &mut tiny_http::Request) -> RouteResponse {
             let body = read_body(req);
             resp(200, sharp_library::handle_relaunch_bottle_installer(&body))
         },
+        (Method::Post, "/bottles/apply-font-subs") => {
+            let body = read_body(req);
```
PR-Reviewer-Ant, line 702: The PR description says post-wineboot config seeding is completed, but this only adds a manual `/bottles/seed-post-wineboot` endpoint. I don't see the seed called from bottle creation, Steam bottle preparation, or the wineboot path, so newly-created prefixes will not automatically get the font substitutions, DLL overrides, or `y:` drive mapping described by the PR. If this is meant to be foundational default behavior, it needs to be wired into the post-wineboot flow rather than left as an endpoint nobody calls.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `fec634ad`

</details>

I found two runtime-behavior issues: the new GPTK install step can report success while installing nothing, and the post-wineboot seeding described as completed is only exposed as a manual endpoint. Both undercut core claims in the PR description.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The installer control flow is explicit in app/src-rust/src/installer.rs: install_gptk_runtime returns Ok(false) when no GPTK archive was found or extraction did not produce required files, and run_install_all reports Ok(false) as a completed ready step. The post-wineboot seed functions and routes are present in bottles.rs/main.rs, but the diff shows no integration into bottle or prefix creation; I did not have additional unchanged call-site context beyond the provided diff.

app/src-rust/src/installer.rs
```diff
@@ -504,6 +506,65 @@ fn install_dxmt_runtime(home: &PathBuf) -> Result<bool, String> {
...
+        Ok(true)
+    } else {
```
PR-Reviewer-Ant, line 566: This makes the new `GPTK D3DMetal Runtime` setup step look successful even when nothing was installed. If no bundled/downloaded `gptk.tar.zst` is found, the function falls through and returns `Ok(false)`, and `run_install_all()` reports the step as `done` / `ready`. That contradicts the PR's auto-install/bundled GPTK claim and leaves M13 failing later with a missing required runtime. This should return an error when GPTK is absent, or at least surface a distinct skipped/unavailable state that setup does not treat as ready.

app/src-rust/src/bottles.rs
```diff
@@ -1860,7 +2060,22 @@ fn inspect_component_state(prefix: &Path, id: &str, fallback: ComponentState) ->
             }
         },
         "vcrun2019" => {
-            if system32.join("vcruntime140.dll").exists() || syswow64.join("vcruntime140.dll").exists() {
+            if system32.join("vcruntime140.dll").exists()
```
PR-Reviewer-Ant: `vcrun2019` is now considered installed if any one of `vcruntime140.dll`, `vcruntime140_1.dll`, or `msvcp140.dll` exists. That can mark a partially installed VC++ runtime as complete and cause `install-recipe-deps` / runtime doctor to skip repairing the component even though games still need the missing DLLs. Since this PR expands dependency automation, readiness needs to require the core DLL set for the relevant arch or return `NeedsRepair` for partial installs.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f6439f7f`

</details>

I found two runtime correctness issues: the GPTK installer can report success without installing GPTK, and VC++ 2015-2022 detection can mark a partial runtime as complete. Both directly undermine the runtime readiness claims in the PR description.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The GPTK setup behavior is explicit in installer.rs: missing archives fall through to Ok(false), which the install-all loop records as a ready step. The VC++ detection issue is also directly visible in bottles.rs, where any one v14 DLL is enough to report vcrun2019 Installed.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

app/src-rust/src/bottles.rs
```diff
@@ -2425,6 +2665,160 @@ fn run_wine_reg_set_windows_version(
...
+            let _ = std::os::unix::fs::symlink(&home, &y_link);
+        }
+    }
+
+    let mut reg = build_font_substitution_reg();
+    reg.push_str("\r\n[HKEY_CURRENT_USER\\Software\\Wine\\DllOverrides]\r\n");
+    for (dll, mode) in POST_WINEBOOT_DLL_OVERRIDES {
+        reg.push_str(&format!("\"{}\"=\"{}\"\r\n", dll, mode));
+    }
+
```
PR-Reviewer-Ant, line 2907: This writes the `metalsharp-post-wineboot-seeded` marker immediately after spawning `regedit`, before the registry import has exited or reported success. If Wine/regedit fails, times out, or exits non-zero, future `prepare_bottle()` calls will skip seeding because the marker already exists. That makes the new post-wineboot defaults non-idempotent and can permanently leave a prefix without the claimed font substitutions, DLL overrides, or drive mapping unless the user manually removes the marker. Please wait for the child and only write the marker after a successful exit, or do not use the marker as proof of success.

app/src-rust/src/bottles.rs
```diff
@@ -1860,7 +2068,23 @@ fn inspect_component_state(prefix: &Path, id: &str, fallback: ComponentState) ->
             }
         },
         "vcrun2019" => {
-            if system32.join("vcruntime140.dll").exists() || syswow64.join("vcruntime140.dll").exists() {
+            let has = |dll: &str| -> bool { system32.join(dll).exists() || syswow64.join(dll).exists() };
+            let core = ["vcruntime140.dll", "vcruntime140_1.dll", "msvcp140.dll"];
+            let core_count = core.iter().filter(|dll| has(dll)).count();
+            if core_count == core.len() {
```
PR-Reviewer-Ant, line 2172: VC++ 2013 is marked installed when either `msvcr120.dll` or `msvcp120.dll` exists. The component description and recipe installer both treat this as the VC++ 2013 runtime, which needs both DLLs; with the current check, a partially installed prefix can be reported as satisfied and `/steam/install-recipe-deps` will skip repair. This undermines the PR's dependency automation for titles that need the C++ runtime. Please require the expected DLL set, or return `NeedsRepair` when only part of it is present, similar to the new `vcrun2019` handling.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `222b6ff1`

</details>

I found two correctness issues in the new runtime seeding and recipe dependency checks. Both can make the new automation report success while leaving prefixes only partially configured.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The issues are directly visible in the changed Rust code: the post-wineboot marker is written immediately after spawning regedit, and VC++ 2013 dependency detection treats a single DLL as a complete install even though the component is defined as requiring both msvcr120 and msvcp120.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

app/src-rust/src/bottles.rs
```diff
@@ -1489,9 +1602,16 @@ fn runtime_profile_definition(profile: RuntimeProfile) -> RuntimeProfileDefiniti
             "D3D12 Metal",
             BottleArch::Win64,
             true,
-            &["d3d12", "d3d11", "dxgi", "vcrun2019"][..],
+            &["d3d12", "d3d11", "dxgi", "vcrun2019", "gpu_vendor_stubs"][..],
```
PR-Reviewer-Ant, line 1703: This makes `gpu_vendor_stubs` a required M12 runtime component, but the component is only detectable and described later; I don't see a matching installer/repair branch that deploys `nvapi64.dll`, `nvngx.dll`, or `atidxx64.dll` into the bottle prefix. That means `/bottles/prepare` or runtime doctor can report a required repair that `repair_component` cannot satisfy, and the new `/steam/install-recipe-deps` path will fail for any recipe/profile that depends on it. The DXMT/GPTK tarballs may contain these DLLs for WINEDLLPATH, but the bottle component contract says the prefix component itself is installable.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `766a7672`

</details>

I found one runtime-install correctness issue: the new GPU stub component is advertised and required by M12/M13 profiles, but the repair/install path does not actually install it.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff shows gpu_vendor_stubs added to runtime profile requirements and diagnostic/action metadata, but resolve_component_installer_from_roots only adds installer resolution for vcrun2013 and existing redistributables. I did not have the full unchanged repair_component implementation in the prompt, so the confidence is medium, but the changed code presented here does not provide an install source for this new component.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/bottles.rs
```diff
@@ -1922,6 +2203,13 @@ fn inspect_component_state(prefix: &Path, id: &str, fallback: ComponentState) ->
                 ComponentState::Missing
             }
         },
```
PR-Reviewer-Ant, line 2248: This marks `gpu_vendor_stubs` installed if either `nvapi64.dll` or `nvngx.dll` exists. The repair path and M12/M13 profiles treat this as a multi-file component, so a prefix with only one stub can be reported ready and skip repair even though the other compatibility DLL is still missing. Please require the expected stub set here, and make the repair path only count a stub after `fs::copy` succeeds.

app/src-rust/src/mtsp/rules.rs
```diff
@@ -45,30 +75,60 @@ fn rule_candidates(home: &Path, current_exe: Option<&Path>) -> Vec<PathBuf> {
...
+            .get("env")
+            .and_then(|v| v.as_table())
+            .map(|t| t.iter().filter_map(|(k, v)| v.as_str().map(|s| (k.clone(), s.to_string()))).collect())
+            .unwrap_or_default();
+        let check_dlls = entry
+            .get("diagnostics")
+            .and_then(|d| d.get("check_dlls"))
+            .and_then(|v| v.as_array())
+            .map(|arr| arr.iter().filter_map(|v| v.as_str().map(String::from)).collect())
+            .unwrap_or_default();
```
PR-Reviewer-Ant, line 122: The PR description says per-game recipes include env and diagnostic checks, and this parser now loads `env` plus `diagnostics.check_dlls`, but the launch and doctor paths in this diff never apply those env vars or check those DLLs. That makes the new TOML fields look supported while they are effectively inert. Either wire them into launch/runtime-doctor behavior or keep the schema narrower until those fields are actually consumed.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `552f9a5e`

</details>

I found a runtime-state bug in the new GPU stub component: it can report the component installed even when only one stub is present or copies failed. There is also some scope drift in the per-game recipe parser: env and diagnostics are parsed but not actually applied or checked.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The GPU stub state bug is directly visible in bottles.rs: repair ignores fs::copy failures and inspect_component_state treats either nvapi64.dll or nvngx.dll as installed. The recipe env/diagnostics drift is also visible from rules.rs parsing those fields while the changed launch and doctor paths only consume dependency lists and recipe_name.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `4d761e3f`

</details>

I found one data-integrity issue in the legacy DLL cleanup path. The new migration code trusts paths from a manifest in the game directory without canonicalizing them, so crafted `..` paths can restore files outside the intended game tree.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The issue is directly visible in the new `cleanup_legacy_injections` implementation: it accepts `dest_path` and `backup_path` from `.metalsharp/injections.json`, checks them with lexical `PathBuf::starts_with`, then copies to the destination. The supplied tests cover an absolute outside path, but not paths containing `..`, which Rust path prefix checks do not normalize away.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/mtsp/launcher.rs
```diff
@@ -693,6 +730,52 @@ fn build_dyld(ms_root: &PathBuf, paths: &[&str]) -> String {
...
+        },
+    };
+
+    if let Some(dlls) = manifest.get("dlls").and_then(|d| d.as_array()) {
+        for dll in dlls {
+            if let Some(backup) = dll.get("backup_path").and_then(|b| b.as_str()) {
+                let backup_path = PathBuf::from(backup);
+                if !backup_path.starts_with(game_dir) {
+                    continue;
+                }
```
PR-Reviewer-Ant, line 789: `cleanup_legacy_injections` can still write outside `game_dir` if the manifest contains a path like `<game_dir>/../outside.dll`. `PathBuf::starts_with` is a lexical component check; it does not canonicalize `..` before this copy. Since the manifest lives under a game directory and this cleanup runs automatically for WINEDLLPATH-routed launches, a corrupted or crafted `.metalsharp/injections.json` can make MetalSharp overwrite files outside the game install. Please canonicalize the game dir, backup path, and destination path before comparing, and only copy when both canonical paths remain under the canonical game dir. If the destination does not exist yet, canonicalize the parent and validate the final file name separately.

app/src-rust/src/bottles.rs
```diff
@@ -1410,6 +1558,44 @@ pub fn handle_steam_compatdata(body: &serde_json::Map<String, Value>) -> Value {
...
+    };
+    let prefix = PathBuf::from(&manifest.prefix_path);
+    let missing = crate::mtsp::rules::game_missing_dependencies(appid, &prefix);
+    if missing.is_empty() {
+        return json!({"ok": true, "appid": appid, "installed": [], "message": "all recipe dependencies satisfied"});
+    }
+
+    let mut installed = Vec::new();
+    let mut errors = Vec::new();
+    for component_id in &missing {
```
PR-Reviewer-Ant, line 1593: `handle_install_recipe_deps` records every successful `repair_component` call as installed, but `repair_component` can return `Ok` for non-installed states. The new `gpu_vendor_stubs` branch, for example, returns `Ok` with `status: "asset_missing"` when the DLLs are unavailable, and normal redistributable installers often return a spawned process report rather than proof that the component is now present. This means `/steam/install-recipe-deps` can return `ok: true` and `message: installed N components` while the recipe dependencies are still missing. Re-check the component state after each repair, or only count reports whose status actually means the component is present.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `b6c478fb`

</details>

I found one correctness issue in the new dependency installer path: it can report successful installation while leaving recipe components missing. That undercuts the new `/steam/install-recipe-deps` behavior for common VC++/DirectX repair cases.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The new endpoint loops over recipe components and treats any Ok(ComponentRepairReport) from repair_component as installed, but repair_component has existing non-error statuses such as asset_missing/pid-based async installs. I can prove the status is ignored from the changed bottles.rs flow, though the exact runtime impact depends on existing repair_component behavior outside the shown diff.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `987fdd52`

</details>

I found one data-loss bug in the legacy DLL cleanup path. The runtime work is broad, but this cleanup routine can remove the only restoration metadata even when restoration did not actually happen.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The issue is directly visible in the new cleanup_legacy_injections implementation: skipped or failed restore entries are ignored, then the entire .metalsharp injection directory is deleted unconditionally. No runtime trace is needed to prove the loss of backup/manifest state.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/mtsp/launcher.rs
```diff
@@ -693,6 +730,66 @@ fn build_dyld(ms_root: &PathBuf, paths: &[&str]) -> String {
...
+                        continue;
+                    }
+                    let _ = std::fs::copy(&canonical_backup, &canonical_dest);
+                    let _ = std::fs::remove_file(&canonical_backup);
+                }
+            }
+        }
+    }
+
+    let _ = std::fs::remove_dir_all(&injection_dir);
```
PR-Reviewer-Ant: This deletes the legacy injection directory even when individual DLL restores were skipped or failed. For example, if a backup path no longer canonicalizes, a dest parent is missing, or `copy` fails, the loop just continues/ignores the error and then removes `.metalsharp`, including the manifest and any remaining backups. That leaves the injected DLL in the game directory with no recovery metadata, which is exactly the state this migration is trying to clean up. Please only remove the directory after all manifest entries that need restoring have been restored successfully, or keep the manifest/backups and report a repair-needed state for partial cleanup.

app/src-rust/src/bottles.rs
```diff
@@ -1147,6 +1165,61 @@ pub fn repair_component(
         });
     }
 
+    if component_id == "gpu_vendor_stubs" {
+        let home = dirs::home_dir().unwrap_or_default();
+        let ms_root = home.join(".metalsharp").join("runtime").join("wine");
+        let system32 = prefix.join("drive_c").join("windows").join("system32");
+        fs::create_dir_all(&system32)?;
+
+        let dxmt_dir = ms_root.join("lib").join("dxmt").join("x86_64-windows");
```
PR-Reviewer-Ant, line 1174: This new `gpu_vendor_stubs` repair path only deploys `nvapi64.dll` and `nvngx.dll`, but the PR claims AMD stub coverage and the M13 pipeline declares `atidxx64.dll` as part of its GPTK DLL set. Because `inspect_component_state` also only checks the two NVIDIA stubs, a bottle can be marked ready while the AMD compatibility stub is absent. Please include `atidxx64.dll` in the component install/verification path or narrow the advertised component so runtime doctor does not over-report readiness.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `7abe0cc2`

</details>

I found one implementation gap: the new GPU vendor stub component claims AMD coverage but never installs or verifies the AMD stub, so M13/GPTK readiness can be reported even when atidxx64.dll is missing.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description explicitly includes GPU vendor stubs for nvapi64, nvngx, and atidxx64. The changed M13 pipeline declares atidxx64.dll, but the new gpu_vendor_stubs repair and inspection paths in bottles.rs only handle nvapi64.dll and nvngx.dll.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/bottles.rs
```diff
@@ -1874,12 +2191,14 @@ fn inspect_component_state(prefix: &Path, id: &str, fallback: ComponentState) ->
             }
         },
         "directx_jun2010" => {
-            if system32.join("d3dx9_43.dll").exists()
-                || syswow64.join("d3dx9_43.dll").exists()
-                || system32.join("xinput1_3.dll").exists()
-                || syswow64.join("xinput1_3.dll").exists()
-            {
+            let has = |dll: &str| -> bool { system32.join(dll).exists() || syswow64.join(dll).exists() };
+            let core_dlls = ["d3dx9_43.dll", "d3dx10_43.dll", "d3dx11_43.dll", "xinput1_3.dll"];
```
PR-Reviewer-Ant, line 2201: `extended_dlls` is introduced but never used. With the repository's `cargo clippy --all-targets -- -D warnings` gate, this becomes a hard failure, so the PR's claimed clippy pass does not line up with the diff.

app/src-rust/src/bottles.rs
```diff
@@ -2425,6 +2763,164 @@ fn run_wine_reg_set_windows_version(
...
+        .env("WINEDEBUG", "-all")
+        .env("WINEDEBUGGER", "none")
+        .stdout(Stdio::from(stdout))
+        .stderr(Stdio::from(log));
+    crate::platform::set_runtime_library_env(&mut cmd, &ms_root);
+    let mut child = cmd.spawn()?;
+    let pid = child.id();
+    let exit_status = child.wait()?;
+    if exit_status.success() {
+        let marker = prefix.join("drive_c").join("metalsharp-post-wineboot-seeded");
```
PR-Reviewer-Ant, line 2930: `fs::write(&marker, timestamp_secs())` passes the numeric timestamp directly as the file contents. `std::fs::write` requires something implementing `AsRef<[u8]>`, and the surrounding uses of `timestamp_secs()` assign it to manifest timestamp fields, so this should fail to compile unless the helper's type changed elsewhere. Convert it to a string/bytes before writing the marker.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `57548980`

</details>

I found two compile-time regressions in the new bottle seeding/component detection path, so the stated Rust verification cannot be true for this diff as shown.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both issues are directly visible in the changed Rust code: an unused local in bottles.rs and a u64 passed to fs::write. These are compile or clippy failures in the touched files.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `678b0445`

</details>

I found one runtime installation bug: the DXMT installer now decides an existing install is stale when `nvapi64.dll` is missing, but it can still report success without actually installing that new required stub.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed readiness check in `install_dxmt_runtime` requires `nvapi64.dll`, while the later success check still only requires `winemetal.so` and `d3d11.dll`. That mismatch is directly visible in the provided `installer.rs` diff and affects the new GPU-stub/runtime readiness behavior claimed by the PR.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/installer.rs
```diff
@@ -445,8 +446,9 @@ fn install_dxmt_runtime(home: &PathBuf) -> Result<bool, String> {
     let dxmt_dir = home.join(".metalsharp").join("runtime").join("wine").join("lib").join("dxmt");
     let unix_so = dxmt_dir.join("x86_64-unix").join("winemetal.so");
     let pe_dll = dxmt_dir.join("x86_64-windows").join("d3d11.dll");
+    let nvapi = dxmt_dir.join("x86_64-windows").join("nvapi64.dll");
 
-    if unix_so.exists() && pe_dll.exists() {
+    if unix_so.exists() && pe_dll.exists() && nvapi.exists() {
         return Ok(false);
     }
```
PR-Reviewer-Ant, line 452: This makes older DXMT installs look stale when `nvapi64.dll` is missing, but the function can still return `Ok(true)` without installing the new stub. If `dxmt.tar.zst` is unavailable and the fallback `~/metalsharp/runtime/dxmt` has the old `winemetal.so` + `d3d11.dll` layout, the final success check passes even though `nvapi64.dll` is still absent. That means `/setup/install-all` can report DXMT ready while the new M9-M12 pipelines and `gpu_vendor_stubs` component expect the stub to exist. The final readiness check should include the same required files as the initial check, or the function should return a partial/missing-asset error when the new stub was not installed.

app/src-rust/src/installer.rs
```diff
@@ -504,6 +505,65 @@ fn install_dxmt_runtime(home: &PathBuf) -> Result<bool, String> {
...
+                } else {
+                    let _ = fs::copy(entry.path(), &dst);
+                }
+            }
+        }
+
+        let _ = fs::remove_dir_all(&tmp);
+    }
+
+    if gptk_dir.join("x86_64-windows").join("d3d11.dll").exists() && framework.exists() {
```
PR-Reviewer-Ant: This success check is too shallow for the new GPTK runtime. `copy_dir_recursive()` ignores all nested copy failures, and this only verifies that `D3DMetal.framework` exists as a directory plus `d3d11.dll` exists. A partial framework copy can still return `Ok(true)`, but M13 later depends on the framework contents being loadable. Please verify specific required files such as `D3DMetal.framework/Versions/A/D3DMetal` and the Resources dylibs, and make the recursive copy return errors instead of silently dropping them.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `1cb18804`

</details>

I found one runtime packaging issue that can make the new M13/GPTK install path report success while leaving required framework contents missing. The rest of the diff broadly matches the PR description, but this should be tightened before relying on bundled GPTK as foundational runtime work.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The installer only verifies the D3DMetal.framework directory exists after a recursive copy that ignores copy errors. I did not have a runtime extraction trace for the actual gptk.tar.zst bundle, so confidence is medium rather than high.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `368b2c04`

</details>

I found one correctness issue in the new recipe dependency path: it can report dependencies as satisfied even when the component detector would still mark them incomplete. That undercuts the new runtime-doctor/install-recipe-deps behavior this PR is adding.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed code in mtsp/rules.rs uses a separate, weaker DLL heuristic for recipe dependencies than the component inspection added in bottles.rs. The mismatch is visible directly in the diff for vcrun2019 and directx_jun2010.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/mtsp/rules.rs
```diff
@@ -100,6 +160,38 @@ pub fn resolve_pipeline(appid: u32) -> PipelineId {
...
+    let system32 = prefix.join("drive_c/windows/system32");
+    let syswow64 = prefix.join("drive_c/windows/syswow64");
+    let component_dll_map: &[(&str, &[&str])] = &[
+        ("vcrun2019", &["vcruntime140.dll"]),
+        ("vcrun2013", &["msvcr120.dll"]),
+        ("directx_jun2010", &["d3dx9_43.dll", "xinput1_3.dll"]),
+        ("corefonts", &["arial.ttf"]),
+    ];
+    let mut missing = Vec::new();
+    for component_id in &recipe.components {
```
PR-Reviewer-Ant, line 176: `game_missing_dependencies` treats a component as present if *any* listed DLL exists, so `POST /steam/install-recipe-deps` and runtime doctor can skip installs while the bottle component is still incomplete. For example `vcrun2019` only checks `vcruntime140.dll`, but `inspect_component_state` now requires `vcruntime140.dll`, `vcruntime140_1.dll`, and `msvcp140.dll`; `directx_jun2010` is similarly satisfied by either `d3dx9_43.dll` or `xinput1_3.dll` even though the component detector now requires the broader core set. This makes the new recipe dependency endpoint report success for partially installed runtimes. It should reuse the same component inspection logic or require the same DLL set here.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `1613827e`

</details>

I found one runtime drift issue: the new recipe dependency endpoint depends on updated mtsp rules, but existing installs keep using stale cached rules and never receive the new dependency metadata.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The changed installer returns early when ~/.metalsharp/configs/mtsp-rules.toml already exists, while the new dependency logic reads recipes from the installed rules file. I did not run an upgrade scenario, but the failure follows directly from install_mtsp_rules and the new recipe parsing path.


Unmapped findings (not on changed lines):
- app/src-rust/src/installer.rs:723 - This preserves an existing installed `mtsp-rules.toml`, which means current users will not get the new `[dependencies]`, `[diagnostics]`, or recipe env fields added in this PR. The new `/steam/install-recipe-deps` and runtime-doctor recipe reporting paths only work if `mtsp-rules.toml` contains those fields, so upgraded installs can report no missing recipe dependencies even for games this PR explicitly declares. The installer needs to update/merge the packaged rules when the bundled rules change, or the new recipe system is only available on fresh installs.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/bottles.rs
```diff
@@ -1489,9 +1694,16 @@ fn runtime_profile_definition(profile: RuntimeProfile) -> RuntimeProfileDefiniti
             "D3D12 Metal",
             BottleArch::Win64,
             true,
-            &["d3d12", "d3d11", "dxgi", "vcrun2019"][..],
+            &["d3d12", "d3d11", "dxgi", "vcrun2019", "gpu_vendor_stubs"][..],
```
PR-Reviewer-Ant, line 1703: M12 now requires `gpu_vendor_stubs`, but that component cannot be satisfied from a DXMT-only runtime. `repair_component("gpu_vendor_stubs")` treats `atidxx64.dll` as GPTK-only and skips it when `lib/gptk/x86_64-windows/atidxx64.dll` is absent, while `inspect_component_state` and `recipe_component_satisfied` require all three DLLs (`nvapi64.dll`, `nvngx.dll`, and `atidxx64.dll`) to mark the component installed. That means a normal DXMT/M12 setup can copy the DXMT NVIDIA stubs and still remain `NeedsRepair`/`still_missing` forever unless GPTK is also installed. Either make the M12 requirement match the DXMT assets, provide the AMD stub in DXMT, or split the GPTK AMD stub into a separate component.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `6e20635c`

</details>

I found one correctness issue in the new GPU vendor stub component: the M12/DXMT path now requires an AMD GPTK-only DLL, which can leave DXMT bottles permanently needing repair even when the DXMT runtime is otherwise complete.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new M12 runtime profile requires gpu_vendor_stubs, repair_component only sources atidxx64.dll from the GPTK directory, and the component is considered installed only when nvapi64.dll, nvngx.dll, and atidxx64.dll are all present. That interaction is directly visible in bottles.rs.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `5a8235c6`

</details>

I found one maintainability/correctness issue worth discussing: the bundled roadmap document appears stale for this PR and describes implementation details that no longer match the code being added.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The mismatch is directly visible in the new roadmap file and the implementation diff, but I did not have a live runtime/bundle artifact to verify the GPTK packaging behavior end to end.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

docs/crossover-parity-roadmap.md
```diff
@@ -0,0 +1,730 @@
+# MetalSharp CrossOver Parity Roadmap
+
+Status: **In Progress** — PR #114 (codex/crossover-parity)
```
PR-Reviewer-Ant: This new roadmap is committed as if it belongs to PR #114 and references an older current-main/version snapshot. It also documents GPTK install behavior that does not match this PR's code: the roadmap says the installer should fetch `gptk-bundle.tar.zst`, install under `lib/gptk/external`, and maybe use a bundled GPTK wine64, while the actual installer looks for `gptk.tar.zst`, installs `D3DMetal.framework` under `runtime/wine/lib/external`, and continues using MetalSharp's Wine plus `WINEDLLPATH`. That kind of stale implementation roadmap is likely to mislead the next runtime maintainer about what Beta 7 actually shipped.

app/src-rust/src/installer.rs
```diff
@@ -504,6 +505,90 @@ fn install_dxmt_runtime(home: &PathBuf) -> Result<bool, String> {
...
+        }
+
+        let _ = fs::remove_dir_all(&tmp);
+    }
+
+    if gptk_runtime_ready(&gptk_dir, &framework) {
+        Ok(true)
+    } else {
+        Err("GPTK D3DMetal runtime incomplete — bundle gptk.tar.zst with required PE DLLs and D3DMetal.framework contents".into())
+    }
```
PR-Reviewer-Ant, line 568: This readiness check can pass with an incomplete GPTK bundle because it requires `nvapi64.dll` and `atidxx64.dll` but not `nvngx.dll`. The PR claims GPTK ships all three vendor stubs, `RuntimeProfile::M13` depends on `gpu_vendor_stubs`, and `repair_component("gpu_vendor_stubs")` later requires both `nvapi64.dll` and `nvngx.dll`. As written, `/setup/install-all` can mark the GPTK runtime ready while the M13 bottle is still missing a required component. Please include `nvngx.dll` in `required_pe` so the setup gate matches the runtime/profile checks.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `abf74b4e`

</details>

I found one concrete gap in the GPTK runtime validation: setup can report the bundled GPTK runtime ready even when one of the PR's claimed required vendor stubs is missing. That undercuts the new M13/GPTK path and the GPU stub work described in the PR.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description and roadmap say GPTK bundles nvapi64, nvngx, and atidxx64, and M13/GPTK profiles now depend on GPU vendor stubs. The readiness gate in app/src-rust/src/installer.rs explicitly omits nvngx.dll from required_pe, so install-all can accept an incomplete GPTK payload before runtime doctor/repair later marks gpu_vendor_stubs missing.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

app/src-rust/src/installer.rs
```diff
@@ -613,11 +694,19 @@ fn install_mtsp_rules(home: &PathBuf) -> Result<bool, String> {
     for src in &candidates {
```
PR-Reviewer-Ant, line 694: This now overwrites an existing `~/.metalsharp/configs/mtsp-rules.toml` whenever the packaged rules differ. That fixes stale bundled recipes, but it also means any user-local pipeline overrides in that file are silently replaced during `install-all`; the single `.bak` helps recovery but does not preserve/merge the active config. If this file is user-editable, consider separating packaged recipes from user overrides or only refreshing files that carry a MetalSharp-managed marker/version.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `2c9813cc`

</details>

I found one correctness risk in the runtime installer flow: the new setup step now overwrites the installed MTSP rules file, which can silently discard local overrides.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff shows install_mtsp_rules no longer returns early when ~/.metalsharp/configs/mtsp-rules.toml exists and now rewrites it from packaged repo candidates. I do not have enough surrounding product context to know whether that installed file is considered user-editable, so this is a warning rather than a blocker.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

tools/dmg/create-bundles.sh
```diff
@@ -16,6 +16,8 @@ mkdir -p "$BUNDLE_DIR"
 BUNDLES=(
     "metalsharp_bundle.tar.zst"
     "metalsharp_bundle2.tar.zst"
+    "dxmt.tar.zst"
+    "gptk.tar.zst"
```
PR-Reviewer-Ant, line 20: This adds `gptk.tar.zst` to the fetched bundle set, but the Electron packaging config shown in `app/package.json` does not add `bundles/gptk.tar.zst` or `bundles/dxmt.tar.zst` to `extraResources`. That means a packaged app will not actually carry the new GPTK/DXMT archives even though `install_gptk_runtime()` expects to find `resources/bundles/gptk.tar.zst` before falling back to a GitHub download. For the PR's “GPTK 3.0 bundled” and auto-install claims, this needs to be wired into the packaged resources, not just the fetch script.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f5bb0a73`

</details>

I found one packaging gap that undercuts the PR's bundled-runtime claim: the new GPTK/DXMT archives are fetched for bundle creation but are not included in Electron's packaged resources, so a packaged app can still fall back to network download or fail setup.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The diff adds dxmt.tar.zst and gptk.tar.zst to tools/dmg/create-bundles.sh, and install_gptk_runtime/find_bundled_archive look for resources/bundles/gptk.tar.zst at runtime. The provided package.json content still only lists metalsharp_bundle.tar.zst, metalsharp_bundle2.tar.zst, and other older bundle files in extraResources, with no dxmt.tar.zst or gptk.tar.zst entries. I did not run an Electron package build here, so confidence is medium rather than high.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR #116, opened 2026-05-23
Fix Beta 7 install and migration foundations

aaf2tbz:
## Summary
- add a runtime bundle preflight step so clean installs require every Beta 7 bundle before extraction begins
- include the uploaded Beta 7 support archives in macOS/Linux bundle fetchers and packaged app resources
- bump migration schema and expand runtime readiness checks so existing installs repair missing Beta 7 foundation files

## Root cause
The clean install wizard could reach the install steps without all of the newly uploaded runtime archives available in packaged resources or cache. Existing installs could also be treated as migration-ready because the migration readiness check only covered the older base runtime surface, not the newer GPTK/DXMT/vendor/support assets.

## Validation
- verified live `bundles` release assets with `gh release view bundles --repo aaf2tbz/metalsharp --json assets,tagName,url`
- `cargo fmt --check --manifest-path app/src-rust/Cargo.toml`
- `cargo test --manifest-path app/src-rust/Cargo.toml`
- `cargo clippy --manifest-path app/src-rust/Cargo.toml --all-targets -- -D warnings`
- `npm run build`
- `git diff --check`


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm taking a look at the fixes in `Fix Beta 7 install and migration foundations` (commit `9f79fa40`) and will follow up shortly.

app/package.json
```diff
@@ -183,6 +183,34 @@
         "from": "bundles/metalsharp_bundle2.tar.zst",
         "to": "bundles/metalsharp_bundle2.tar.zst"
       },
+      {
```
PR-Reviewer-Ant, line 186: This makes Linux packaging require macOS-only bundle files that the Linux fetch path does not produce. `tools/linux/create-bundles.sh` now fetches `dxvk`, `mono-x86`, `goldberg`, and `eac-toggle`, but it still does not fetch `dxmt.tar.zst`, `gptk.tar.zst`, or `mono-arm64.tar.zst`; `LINUX_RUNTIME_BUNDLE_ASSETS` also correctly omits them. Because `extraResources` is unconditional, a clean `METALSHARP_TARGET=linux npm run dist`/DEB build can fail when electron-builder tries to copy these missing files, or pass only because stale macOS bundle artifacts are sitting in `app/bundles`. Please split target-specific resources or make the Linux bundle script produce every file listed here.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `9f79fa40`

</details>

I found one packaging regression: the Linux bundle fetcher and Linux runtime preflight intentionally omit macOS-only Beta 7 archives, but electron-builder is now told to package those archives unconditionally. That makes Linux packaging depend on stale local macOS bundle files or fail on a clean checkout.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The mismatch is visible directly in this diff: app/package.json adds dxmt.tar.zst, gptk.tar.zst, and mono-arm64.tar.zst to unconditional extraResources, while tools/linux/create-bundles.sh and LINUX_RUNTIME_BUNDLE_ASSETS do not fetch or require those files for Linux.

PR #206, opened 2026-06-14
Disable mscompatdb for M12 launches

aaf2tbz:
## Summary
- explicitly disable the legacy `mscompatdb` Wine sidecar for M12 via `WINEDLLOVERRIDES`
- keep the existing M12 native/builtin D3D12/DXMT routing unchanged
- add regression coverage for the engine definition and Steam handoff env

## Root Cause
M12 includes `lib/wine/x86_64-unix` in its runtime route, which leaves Wine free to load the old v1 `mscompatdb.so` TOML hook. That legacy sidecar is not part of the isolated M12 DXMT route and can fail launch before the game reaches the D3D12 path.

## Validation
- `cargo test mtsp::engine::tests::m12_is_stronger_than_other_dxmt_d3d_paths`
- `cargo test mtsp::launcher::tests::steam_pipeline_env_includes_route_overrides_and_cache_keys`
- `cargo test mtsp::engine`
- `cargo test mtsp::launcher`

### DIFF
--- a/app/src-rust/src/mtsp/engine.rs
+++ b/app/src-rust/src/mtsp/engine.rs
@@ -89,7 +89,7 @@ pub fn pipelines() -> &'static Vec<PipelineNode> {
                 experimental: false,
                 requires_wine: true,
                 wine_overrides: Some(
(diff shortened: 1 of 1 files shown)


aaf2tbz:
Superseded by #207, which restores PR #204 onto current main and includes this M12 mscompatdb runtime override in the correct PR context.

PR #212, opened 2026-06-15
fix: install updater DMG before relaunch

aaf2tbz:
## Summary
- replace `/Applications/MetalSharp.app` with the verified app from the mounted update DMG before relaunching
- launch the installed `/Applications` copy instead of running directly from the DMG
- keep Python and shell updater handoff paths in sync, including admin-copy fallback and post-copy verification

## Validation
- `python3 -m py_compile app/updater/update.py`
- Python smoke test for `copy_app_bundle()` using a temporary fake app bundle
- `bash -n app/updater/update.sh`
- `git diff --check`

## Notes
- This is a draft because the full macOS updater flow still needs to be exercised with a real signed DMG.


aaf2tbz:
Added follow-up commit `f631a57`:

- during the migration verifier, after Wine prefix/Steam update checks, verify mounted external drives are linked into `prefix-steam/dosdevices`
- create/verify `c:` and add free drive-letter mappings for external `/Volumes/*` drives while skipping system/MetalSharp DMG volumes
- keep the verifier stage visible for ~5 extra seconds before moving on

Additional validation:
- `cargo test verifier_ -- --nocapture`
- `cargo test discover_external_drive_mounts_skips_system_and_dmg_volumes -- --nocapture`
- `cargo test migration_ready_requires_runtime -- --nocapture`
- `git diff --check`

aaf2tbz:
Added follow-up commit `fd1eb56`:

- before opening the newly installed `/Applications/MetalSharp.app`, run an allow step
- removes recursive `com.apple.quarantine` attributes and adds a local `spctl` label when available
- falls back to an administrator shell script if the non-privileged allow step cannot clear quarantine
- applies the same behavior to both Python and shell updater handoff paths

Validation:
- `python3 -m py_compile app/updater/update.py`
- Python smoke test for `allow_installed_app()`
- `bash -n app/updater/update.sh`
- `git diff --check`

PR #374, opened 2026-08-05
fix: clean up failed Wine prefix migrations

TimDarcet:
## Summary

Prevent a failed or timed-out `wineboot -u` from leaving a Wine prefix half-alive while migration reports success. This was reproduced during a 0.58 upgrade: `wineboot` timed out, orphaned prefix services kept running, and the next Steam launch blocked on `reg import`.

## Changes

- bound `wineboot` and Wine-server control commands with reusable timeout handling
- reap the direct child, send prefix-scoped `wineserver -k`, and verify shutdown with `wineserver -w` after timeout, wait failure, or nonzero exit
- restore Steam-library and custom `dosdevices` links before propagating a failed prefix update
- stop migration at the prefix-update step, retain the post-update marker, and report an error instead of saving successful migration metadata
- add regression coverage for child reaping, exact-prefix cleanup, and the no-running-server case

## PR Readiness (MANDATORY)

- [x] Version triple (`CMakeLists.txt`, `Cargo.toml`, `package.json`, `package-lock.json`) in sync if version bumped

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy --all-targets -- -D warnings` passes
- [x] `cargo build --release` passes
- [ ] `cargo test` passes: 645/646 pass locally; the unrelated `kernel_translation::ipc_bridge::tests::test_nt_query_virtual_memory_basic_returns_success` failure reproduces unchanged on the base branch. Upstream Rust CI passes.
- [x] C++ compiles if changed (not changed)
- [x] `clang-format --dry-run --Werror` passes on changed native files (none)
- [x] `ctest` passes if native tests changed (not applicable)
- [x] TypeScript compiles if changed (not changed)
- [x] Biome + Prettier pass if TS/JS changed (not changed)
- [x] Shell scripts lint if changed (not changed)
- [x] Rules TOML validation if changed (not changed)
- [x] DMG workflow validation if release tooling changed (not changed)
- [x] Tested with The Binding of Isaac: Rebirth using M11(32) from an internal drive after recovering the reproduced timeout state

## Test notes

- `cargo test migrate::tests::` — 43 passed
- focused timeout/prefix cleanup tests pass
- strict Clippy, formatting, and release build pass
- full backend suite locally: 645 passed, 1 unrelated pre-existing IPC test failed and also fails on the base branch
- upstream Rust CI passes
- real failure evidence: migration timed out at 120 seconds with a live prefix `wineserver` and `rundll32 ... wine.inf`; after prefix-scoped cleanup, Steam reached fresh login success and Isaac launched with OpenGL 2.1 / GLSL 1.20

## Risk

The cleanup is scoped by `WINEPREFIX` and uses the selected runtime's own `wineserver`; it does not scan or kill unrelated Wine prefixes. A prefix-update failure now leaves migration incomplete and retains the update marker, favoring an explicit retry over a false ready state. Rollback is a revert of commit `46900d10`, which restores the previous non-fatal behavior.

### DIFF
--- a/app/src-rust/src/migrate.rs
+++ b/app/src-rust/src/migrate.rs
@@ -2,8 +2,9 @@ use serde_json::json;
 use std::cmp::Ordering as CmpOrdering;
 use std::fs;
 use std::path::{Path, PathBuf};
-use std::process::Command;
+use std::process::{Child, Command, ExitStatus};
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::{Duration, Instant};
 
 const MIGRATE_VERSION: &str = env!("CARGO_PKG_VERSION");
 const MIGRATE_SCHEMA_VERSION: u64 = 4;
@@ -57,6 +58,9 @@ const MIGRATION_STEAM_METADATA_DENY_NAMES: &[&str] =
     &["cache", "common", "compatdata", "crashes", "depotcache", "downloading", "logs", "shadercache", "Temp", "tmp"];
 const MIGRATION_TOTAL_STEPS: usize = 8;
 const MIGRATION_PRESERVE_TEMP_PREFIX: &str = "metalsharp-migration-preserve-";
+const WINEBOOT_TIMEOUT: Duration = Duration::from_secs(120);
+const WINE_PROCESS_POLL_INTERVAL: Duration = Duration::from_millis(500);
+const WINESERVER_CLEANUP_TIMEOUT: Duration = Duration::from_secs(10);
 
 static MIGRATING: AtomicBool = AtomicBool::new(false);
 
@@ -573,7 +577,12 @@ fn run_migration() {
     );
     match update_existing_wine_prefixes(&ms_dir, step) {
         Ok(updated) => log_to_file(&format!("Migration: wineboot -u completed for {} prefix(es)", updated)),
-        Err(e) => log_to_file(&format!("Migration: wineboot -u failed (non-fatal): {}", e)),
+        Err(e) => {
+            let message = format!("Wine prefix update failed: {}", e);
+            write_migrate_progress("error", step, total_steps, &message, Some(&e));
+            log_to_file(&format!("Migration stopped: {}", message));
+            return;
+        },
     }
     register_external_steam_libraries(&ms_dir);
     clear_steam_crash_marker(&ms_dir);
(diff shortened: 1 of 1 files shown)


aaf2tbz (approved):
Reviewed by Hermes Agent — approving with minor suggestions (none blocking).

## Code Review Summary

**Verdict: Approved** (verified locally: 646/646 tests pass incl. all 3 new regression tests, fmt + clippy clean, release build passes, applies cleanly to current main)

### What's good
- **Correct root-cause fix**: bounded `wineboot -u` wait (120s / 500ms poll) replaces the fixed 240-attempt loop; on timeout/nonzero-exit the direct child is killed+reaped and the prefix is cleaned via `wineserver -k` then verified with `wineserver -w` — verified against the real wineserver semantics (`-k` kills, `-w` waits).
- **Prefix-scoped cleanup**: `WINEPREFIX` + `set_runtime_library_env` on the wineserver invocation means only the failing prefix is touched — consistent with the isolation contract from #372 (never kill another prefix's processes).
- **dosdevices restore before error propagation** (`update_existing_wine_prefixes`): the old `?` early-return skipped the restore on failure; now the snapshot is restored before the error propagates. Genuine fix.
- **Marker retained on early return**: the post-update marker is only removed on the success path, so a failed prefix update re-runs migration next launch (self-healing) instead of silently marking migration complete.
- **Clean error chaining**: cleanup failures are appended to the original error (`wineboot_failure_after_cleanup`).

### Suggestions (non-blocking)
- **`stop_prefix_wine_processes`**: if `-k` fails but `-w` succeeds the kill error is dropped (returning Ok). Intentional per the no-running-server test, but consider logging the `-k` failure at debug level so a genuinely odd kill failure isn't invisible.
- **Multi-prefix loop is now all-or-nothing**: a failure on prefix-steam stops the loop, so the GOG bottle prefix never updates on that run. The retained marker retries next launch, so it self-heals — but a persistently broken prefix-steam now also blocks GOG updates. Consider `continue`ing to process remaining prefixes, collecting errors, then failing at the end — same stop+cleanup guarantee, but doesn't starve unrelated prefixes.
- **Mid-progress log dropped**: the old `still running after 60 seconds` line at attempt 120 is gone — a 120s wineboot now logs nothing until the end. Consider re-adding a `log_to_file` after half the timeout for observability.
- **10s `-w` timeout**: if a slow server takes longer to die, cleanup reports failure and the user sees a chained error even though the kill was delivered. Consider one retry of `-w` before failing.

### Looks good
- 3 new tests cover the real behaviors: timeout+reap, exact-prefix scoping (fake wineserver records `WINEPREFIX` + arg), and the no-running-server case.
- Clean diff: one file, +202/−25, focused on migrate.rs, no unrelated churn.

PR #376, opened 2026-08-05
fix: make virtual memory self-queries reliable

TimDarcet:
## Summary

Fix the macOS `NtQueryVirtualMemory` bridge test and the self-process query path it exposed. The test used hardcoded address `0x10000`, which is unmapped on current macOS, and self queries unnecessarily went through entitlement-sensitive `task_for_pid`.

## Changes

- use `mach_task_self()` directly for self-process virtual-memory queries
- retain `task_for_pid` and port deallocation for other processes
- query the address of a live local value instead of an assumed low mapping
- validate the real `RegionSize` field and that the returned region contains the requested address

## PR Readiness (MANDATORY)

- [x] Version triple (`CMakeLists.txt`, `Cargo.toml`, `package.json`, `package-lock.json`) in sync if version bumped

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy --all-targets -- -D warnings` passes
- [x] `cargo build --release` passes
- [x] `cargo test` passes — 643 passed
- [x] C++ compiles if changed (not changed)
- [x] `clang-format --dry-run --Werror` passes on changed native files (none)
- [x] `ctest` passes if native tests changed (not applicable)
- [x] TypeScript compiles if changed (not changed)
- [x] Biome + Prettier pass if TS/JS changed (not changed)
- [x] Shell scripts lint if changed (not changed)
- [x] Rules TOML validation if changed (not changed)
- [x] DMG workflow validation if release tooling changed (not changed)
- [x] Tested with The Binding of Isaac: Rebirth using M11(32) on an internal drive; this change does not alter game launch routing

## Test notes

- `cargo test` — 643 passed, 0 failed
- focused `NtQueryVirtualMemory` test passes on macOS
- strict Clippy, formatting, and release build pass
- The Binding of Isaac: Rebirth was previously verified with M11(32) on an internal drive; no launch behavior changed here

## Risk

Low. The behavioral change only bypasses `task_for_pid` when the target PID is the current process, using the task port already provided by `mach_task_self()` and avoiding deallocation of that borrowed port. Other-process lookup and ownership are unchanged. Rollback is a revert of commit `47edcfa6`.

### DIFF
--- a/app/src-rust/src/kernel_translation/ipc_bridge.rs
+++ b/app/src-rust/src/kernel_translation/ipc_bridge.rs
@@ -215,11 +215,18 @@ mod xnu_vm {
 
     pub fn query_vm_region(pid: u32, base_address: u64) -> Option<VmRegion> {
         unsafe {
-            let mut task: u32 = 0;
-            let kr = task_for_pid(mach_task_self(), pid as i32, &mut task);
-            if kr != KERN_SUCCESS {
-                return None;
-            }
+            let self_task = mach_task_self();
+            // Self queries neither require task_for_pid entitlement nor return an owned port.
+            let (task, deallocate_task) = if pid == libc::getpid() as u32 {
+                (self_task, false)
+            } else {
+                let mut task: u32 = 0;
+                let kr = task_for_pid(self_task, pid as i32, &mut task);
+                if kr != KERN_SUCCESS {
+                    return None;
+                }
+                (task, true)
+            };
 
             let mut addr = if base_address == 0 { 0x1000 } else { base_address };
             let mut size: u64 = 0;
@@ -237,7 +244,9 @@ mod xnu_vm {
                 &mut object_name,
             );
 
-            let _ = mach_port_deallocate(mach_task_self(), task);
+            if deallocate_task {
+                let _ = mach_port_deallocate(self_task, task);
+            }
 
             if kr != KERN_SUCCESS {
                 return None;
@@ -841,12 +850,14 @@ mod tests {
     }
 
     #[test]
-    fn test_nt_query_virtual_memory_basic_returns_success() {
+    fn test_nt_query_virtual_memory_mapped_address_returns_success() {
         let h = alloc_handle(getpid() as u32, 0, 0x1F0FFF, "Process");
+        let marker = 0u64;
+        let base_address = (&marker as *const u64) as usize as u64;
 
         let mut req = Vec::new();
         req.extend_from_slice(&h.to_le_bytes());
-        req.extend_from_slice(&0x10000u64.to_le_bytes());
+        req.extend_from_slice(&base_address.to_le_bytes());
         req.extend_from_slice(&0x00u32.to_le_bytes());
         req.extend_from_slice(&48u32.to_le_bytes());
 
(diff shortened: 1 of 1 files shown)


aaf2tbz (commented):
Reviewed by Hermes Agent — production change looks correct; one test-validity issue to fix.

## Code Review Summary

**Verdict: Comment** — the self-query fix is correct; the new test has a wrong byte offset that makes its RegionSize assertions pass trivially.

### ✅ The fix itself is right
- **`mach_task_self()` for self-queries** avoids entitlement-sensitive `task_for_pid` — correct for the bridge's own-process queries and explains the flaky test on modern macOS.
- **Port ownership handled correctly**: self path returns an unowned port and skips `mach_port_deallocate` (deallocating `mach_task_self()`'s port would be wrong); non-self path retains `task_for_pid` + deallocation.
- **Live-address query** (address of a stack `marker`) is far more robust than the hardcoded `0x10000`, which is unmapped on current macOS.
- Verified: 643/643 tests pass, fmt + clippy clean, merges into current main with 0 conflicts.

### ⚠️ Warning — wrong RegionSize offset in the test
The response serialization in `handle_nt_query_virtual_memory` (line 520-527) is:
- `[0..8]` BaseAddress, `[8..16]` AllocationBase, `[16..20]` AllocationProtect, `[20..24]` padding, **`[24..32]` RegionSize**, `[32..36]` State, `[36..40]` Protect, `[40..44]` Type

But the test reads `region_size = resp[32..40]` — that's **State (4B) + Protect (4B)**, not RegionSize. With MEM_COMMIT(0x1000) + PAGE_READWRITE(0x04), `[32..40]` decodes to `0x400001000` — a huge number, so `region_size > 0` and the contains-check pass *trivially* regardless of the actual region. The PR description says "validate the real RegionSize field" but it doesn't.

Fix: read `resp[24..32]` for region_size. (I verified the offset with a standalone probe reproducing the serialization: [32..40] = state<<32|protect.)

### 💡 Minor
- `region_base` is read from `[8..16]` (AllocationBase); `[0..8]` is the BaseAddress field. Both are populated with `region_base` in this serialization so the assertion holds, but `[0..8]` is the semantically correct field to check.

### ✅ Looks good
- Non-self query path untouched; handle → pid resolution unchanged.
- Test still meaningfully guards the fix: a failing query would return non-SUCCESS and fail the status assertion.

Please fix the `[32..40]` → `[24..32]` offset and I'll re-review.

aaf2tbz:
Please address the one warning in a follow up commit, then this will be ready for merge :)

aaf2tbz:
Correction from the reviewer: my earlier comment flagged `region_size` at `resp[32..40]` as reading State+Protect instead of RegionSize — **that warning was wrong**, and no change is needed.

`handle_nt_query_virtual_memory` serializes into a 48-byte body (`data`) where RegionSize sits at `data[24..32]`, and `ipc_ok` prepends an 8-byte header (status + return_len). So in the response the field lands at `resp[8+24 .. 8+32] = resp[32..40]` — exactly what the test reads. I verified this by replicating the serialization: `resp[32..40] == 0x4000 == RegionSize`, while `resp[24..32]` (my claimed offset) decodes as alloc_protect+padding.

The test's `region_base` (`resp[8..16]` = data[0..8] = BaseAddress) and the containment assertions are also correct. Apologies for the noise — the PR stands as-is.

TimDarcet:
You're so fast I didn't even get the time to answer haha

Thanks for merging!

aaf2tbz:
No worries lol!

PR #377, opened 2026-08-05
feat: M12 moves to vkd3d-proton (D3D12 -> Vulkan -> MoltenVK) with DXMT rollback

aaf2tbz:
## Summary

M12 (D3D12 pipeline) moves from DXMT to the **vkd3d-proton** graphics stack, per the VKMT custom d3d12 implementation: **D3D12 → vkd3d-proton → Vulkan → MoltenVK → Metal**. M12 no longer points at DXMT by default. It deploys `d3d12.dll` + `d3d12core.dll` from vkd3d-proton, `dxgi.dll` from DXVK (vkd3d-proton ships no dxgi of its own — DXVK 2.1+ shares the DXGI implementation), and runs on the VKMT-patched MoltenVK.

DXMT remains available as a rollback (`m12Backend=dxmt` in Settings → M12 Graphics Backend) and keeps powering M9/M10/M11 unchanged.

## What changed

- **Pipeline re-target** (`engine.rs`): M12 resolves to a config-selected node — `vkd3d-proton` (default) or `dxmt` (fallback). Deploy set: d3d12.dll, d3d12core.dll (vkd3d-proton lane), dxgi.dll (DXVK lane), nvapi64/nvngx stubs. Overrides `d3d12,d3d12core,dxgi=n;b`.
- **Launcher env** (`launcher.rs`): vkd3d-proton cache env branch pins `VK_ICD_FILENAMES` to the runtime MoltenVK ICD, sets `VKD3D_SHADER_CACHE_PATH` + `DXVK_STATE_CACHE_PATH`, never DXMT_* keys. `deploy_prefix_route_dlls` now routes `d3d12core.dll` into system32 (was silently skipped — the forwarder needs it).
- **Installer/resolver** (`installer.rs`): `VKD3D_PROTON_EXPECTED_HASHES` pins the VKMT win64-filtered builds; new lane dirs + readiness/currency/artifact-valid helpers for vkd3d-proton, DXVK, VKMT MoltenVK; `ensure_vkd3d_proton_runtime_ready` stages all three from the graphics bundle; setup stages them best-effort (DXMT fallback if absent). MoltenVK resolver prefers the VKMT lane (`moltenvk_vkmt_ready`, `moltenvk_library_path`, ICD staged + rewritten).
- **Migration wizard** (`migrate.rs`): `reconcile_m12_backend` keeps the vkd3d-proton default when lanes staged, otherwise falls back to `m12Backend=dxmt` so the default never points at a missing runtime.
[...]

PR #395, opened 2026-08-09
fix(m12): harden vkd3d-proton migration and protected-game launches

TimDarcet:
## Summary

This follows the core M12/vkd3d-proton migration already merged in #377 and hardens it for production use:

- pin and validate the gameplay-tested x64 vkd3d-proton and VKMT MoltenVK payloads while preserving all five graphics lanes during runtime refresh
- launch protected games through their real executable without replacing game files, with Elden Ring's D3D feature level and shader cache isolated to the M12 route
- make offline Steam emulation recoverable by caching official Steam API files, preserving `.dll.orig`, and repairing modern Goldberg settings and Steam CEF behavior
- reap detached Wine children so completed games do not remain as zombies
- isolate migration/runtime tests from installed application state

## Validation

- `cargo test`: 749 passed
- `cargo clippy --all-targets -- -D warnings`
- `cargo build --release`
- `tools/ci/check-clang-format.sh`
- 0.59.1 migration: schema 4, up to date; all five graphics lanes retained
- M12 dry run: no missing artifacts; launch doctor: `Ready for M12 via vkd3d-proton`, no blockers or warnings
- Elden Ring offline test: title/offline dialogs, menu, save load, Chapel of Anticipation gameplay, movement, jump, attack, camera, System menu, sustained play/menu interval, save, and Return to Desktop
- post-rebase smoke launch remained active and rendering for several minutes, with no device-loss, resource-lifetime, crash, or fatal signature and no backend-child zombie after exit
- Goldberg disable/re-enable restored and retained the official `steam_api64.dll` in both `.dll.orig` and the per-app cache

Validated payloads:

- `d3d12.dll`: `9b9be3098f45ec3921ed316e7c7bc421790d28a31795722153dc3e532d18aedc`
- `d3d12core.dll`: `844b044e570a37028236b1328a12a8efbe606385a007ac0bac402c683d882746`
- `libMoltenVK.dylib`: `fe9b0ec34f7cefd0497e741d83d919ad5a6b912b1e9abea9c4ef1d1531c9392d`

## Release dependency

The published graphics bundle must be rebuilt and released with the pinned payloads above. The previous release bundle intentionally makes readiness fail and falls M12 back to DXMT because its vkd3d-proton/MoltenVK hashes do not match. The VKMT MoltenVK source patch is a separate upstream dependency and is not included here.

## Known limitations

- offline play only; this does not claim EAC or online support
- MoltenVK still cannot execute the game's D3D12 geometry shaders; affected PSOs fail Metal stage-interface validation
- performance in the tested area is roughly one-sixth speed with high CPU use


aaf2tbz:
How in the world did you get Elden Ring to launch offline? I'm thoroughly impressed. I've had so many issues with that. My only resolution has been the renamed protected executable.

aaf2tbz:
Also the bundle was re-uploaded with a new set so that's what caused the hash mismatch, it just needs to know about the new set. If you ever want the source, it is here: https://github.com/metalsharp/MetalSharp/releases/tag/bundles

aaf2tbz:
Now adding geometry shader support. This was the leading cause of failures.

TimDarcet:
Oh wait I'm on it too! Forgot to check my mails

TimDarcet:
There's only 6 geometry shaders in elden ring, my codex translated them manually it looks quite good but I haven't tested throughly.

It's a very ad hoc solution though, if an update brings new shaders the support won't work. The cleaner path would be to implement general support for translating any geometry shaders at the translation layer level (SPIR-V afaiu), but that's a bit more involved, idk if it's realistic to try and do that. If you can get it working, it would be super cool though!

PR #583, opened 2026-09-07
fix(m12): stabilize Elden Ring launch and performance

TimDarcet:
## Summary

Ports the validated Elden Ring M12 stack from `TimDarcet/MetalSharp` to the current C backend. M12 now routes D3D12 through the pinned vkd3d-proton + DXVK + VKMT MoltenVK stack instead of DXMT.

Source branch: `TimDarcet/MetalSharp@codex/metalsharp-elden-ring-20260903` (`b3740765` through `6b26e052`).

## Changes

- Route M12 through vkd3d-proton, DXVK, and the isolated `moltenvk-vkmt` lane.
- Stage only `d3d12.dll`, `d3d12core.dll`, `d3d11.dll`, and `dxgi.dll` beside the game executable.
- Port the validated private-Metal-API, retained-command-buffer, feature-level, cache, and opt-in diagnostic settings.
- Pin the validated TimDarcet component artifact hashes.
- Keep and package the Elden Ring performance-core topology shim.
- Interpose both `sysconf` and `sysctlbyname`, while forwarding unrelated calls without the `RTLD_NEXT` recursion that crashed Wine.
- Reset an orphaned MetalSharp wineserver before direct launch, but never while a Wine client or Windows executable is running.
- When Background Music is the default output, silently prewarm its CoreAudio pass-through before starting Wine so games still route through BGM and retain per-process volume control.
- Update launch diagnostics and add an M12 regression smoke test.
- Record the validated component provenance and hashes.

## PR Readiness (MANDATORY)

- [x] Version metadata (`CMakeLists.txt`, `app/src-c/Makefile`, `package.json`, `package-lock.json`) in sync if version bumped

- [x] C backend builds and targeted tests pass
- [ ] C++ compiles if changed: not changed
- [x] `git diff --check` passes; `clang-format` is not installed locally
- [ ] `ctest --test-dir build-native`: native engine not changed
- [x] TypeScript and Vite production build pass
- [ ] Shell scripts lint if changed: no shell scripts changed
- [x] Rules TOML validation passed
- [x] Electron package build and deep ad-hoc signature verification pass
- [x] Tested with Elden Ring 1245620, direct M12 launch, internal drive

## Test notes

- Universal x86_64/arm64 topology shim builds with `-Wall -Wextra -Werror`.
- x86_64 interposition probes report 10 performance cores through `sysconf`, `hw.logicalcpu`, and `hw.ncpu`, while unrelated page-size forwarding remains 4096.
- Backend smoke, JSON, migration, HTTP, updater, Steam-process, shadPS4, and PCSX2 tests pass. The updater poll remains intermittently flaky in the aggregate run and passes in isolation.
- SharpEmu's sandbox test cannot run here because macOS rejects `sandbox-exec` with `sandbox_apply: Operation not permitted`; this is unrelated to the changed code.
- `npm run build` and `npm run pack` pass; the package passes deep signature verification.
- With Background Music cold, a PID-scoped trace identified Wwise startup blocked during CoreAudio initialization. Prewarming BGM before launch let exact Elden Ring PID 60111 reach the title screen in under one minute while using `winecoreaudio.so`; BGM then reported the same PID with `rvol = 0`, confirming selective mute works.

## Risk

The release runtime and graphics archives must be republished from the tagged `TimDarcet/vkd3d-proton`, `TimDarcet/dxvk`, and `TimDarcet/MoltenVK` artifacts so their bytes match the new manifests. The BGM prewarm runs only when the default CoreAudio output UID is `BGMDevice`; all other audio configurations are unchanged. The old DXMT and standalone VKD3D lanes remain present; rollback is reverting these commits or selecting another pipeline while new bundles are published.


aaf2tbz:
Curious if it got into actual gameplay? Also my only discrepancy with this is that m12 is currently a publicly removed route while I finish dxmt d3d12 fully. If this can be tied to the vkd3d route instead, that would be most optimal. Also because some m11 games such as Meccha Chemeleon use a subset of dxmt d3d12 already-implemented work to run and without these dlls in their normalized paths will not run. As always I appreciate the hard work you did here to get elden ring launching under the new stack.

aaf2tbz:
Last suggestions as well: Currently feature level is set at 12_0 which is reminiscent of the old vkd3d-proton stack. The more updated version supports up to 12_2 / Core 1.0 . Would you be able set feature level to 12_2 , as well as see if it works under the more recently updated vkd3d-proton-MacOS set? https://github.com/metalsharp/VKD3D-Proton-MacOS . If this doesn't work please let me know. I will be more than happy with this pr if it lands under the vkd3d route with the more updated set and you can confirm it launches elden ring this way. I'm trying to keep things at their most recent versions but if the most recent version fails I will swing over to your pinned hashes.

app/src-c/runtime/diagnostics.c
aaf2tbz, line 1: Please revert this back to how it was while keeping this update on the VKD3D lane.

app/src-c/runtime/game.c
aaf2tbz, line 1: Same thing here. Modify the vkd3d lane preferably to get this working.

app/src-c/runtime/migration.c
aaf2tbz, line 1: Here i am a little confused. The updated vkd3d proton set is preffered if possible but I can switch to your branch if it has working changes that the current one does not.

app/src-c/runtime/mtsp.c
aaf2tbz, line 1: Again here it will need to be reverted and brought to the vkd3d lane or other.

app/src-c/runtime/setup.c
aaf2tbz, line 1: Again not positive about this hash yet. Want to see if the current one works first.

app/src-c/tests/smoke.sh
aaf2tbz, line 1: If possible please set max feature level 12_1 or 12_2

src/wine/cpu_topology_interpose.c
aaf2tbz, line 1: Good edition, this is nice.

tools/bundles/dxvk-runtime-hashes.tsv
aaf2tbz, line 1: Confused here as well with those hashes.

tools/bundles/vkd3d-proton-runtime-hashes.tsv
aaf2tbz, line 1: Overall this pr is good and obviously does what it intends to do through the modified m12 route. The only reason I don't want this on m12 is because dxmt d3d12 is being finished and m12 will be usable again at some point. And i'm a little conflicted about the current hashes you have in place. But I will revert to those hashes if the current more updated set doesn't work for this. Great job on the interpose and audio changes.

aaf2tbz (changes requested):

aaf2tbz:
You'll be pleased to know I got gptk 4 d3dmetal loading with the updated wine runtime i'm making and it runs elden ring perfectly, with steam as well. I will probably just go with what you've done here now that I've gotten this figured out, I may make a few adjustments.

TimDarcet:
> Curious if it got into actual gameplay?

Can confirm in terms of gameplay, it'a perfect! There used to be a lot of graphics bugs because of unsupported geometry shaders, but with these changes it's good and it runs smooth at graphics quality max.

Will come back to the other comments later :)

aaf2tbz:
Not sure what to do with this yet so i'm leaving it here for now until I feel good about it. I tried using this pr in the latest version and couldn't get Elden Ring running with it no matter what I tried so I don't know what I'm doing with it yet lol. I appreciate this work though so I don't want to disband it.

PR #616, opened 2026-09-11
fix: render Wine Steam at Retina resolution

TimDarcet:
## Summary

Enable native Retina rendering for Wine Steam instead of letting macOS upscale a 96-DPI backing surface, which makes the Steam client pixelated and blurry.

## Changes

- Add a default-on `retinaMode` setting and Settings toggle.
- Apply Wine `RetinaMode=Y` with `LogPixels=192` before managed launches; disabling the option restores `RetinaMode=N` and 96 DPI.
- Reuse the existing launch-time registry import so both fresh and existing prefixes converge without rebuilding or deleting the shared prefix.
- Document the shared-prefix performance tradeoff and required Wine Steam restart.
- Add config API and generated-registry regression coverage.

## PR Readiness (MANDATORY)

- [x] Version metadata (`CMakeLists.txt`, `app/src-c/Makefile`, `package.json`, `package-lock.json`) in sync if version bumped

- [ ] C backend builds and tests: `make -C app/src-c test`
- [x] C++ compiles if changed: not changed
- [x] `clang-format --dry-run --Werror` passes on changed C files
- [x] `ctest --test-dir build-native` passes if tests changed: native C++ tests not changed
- [x] TypeScript compiles if changed: `cd app && npx tsc --noEmit`
- [x] Biome + Prettier pass if TS/JS changed
- [ ] Shell scripts lint if changed: ShellCheck unavailable locally; CI will run it
- [x] Rules TOML validation: not changed
- [x] DMG workflow validation: release/bundle tooling not changed
- [x] Tested with Satisfactory (App 526870), M11, managed Steam prefix/C: drive

## Test notes

- `runtime-routes-test` passes, including both Retina-enabled and disabled registry output.
- Config smoke coverage verifies the default and persisted opt-out.
- `tsc --noEmit`, Biome, Prettier, and the Vite production build pass. Biome reports five pre-existing CSS specificity warnings.
- Full `make -C app/src-c test` passes through the new tests and all suites up to SharpEmu; the host then rejects SharpEmu's `sandbox-exec` probe with `sandbox_apply: Operation not permitted`.
- Satisfactory launched through M11 and remained running with the shared prefix after Retina mode was enabled, then was stopped through the backend.

## Risk

Retina mode doubles the Wine backing resolution and can increase GPU load for games because they share `prefix-steam`. Users can disable High Resolution in Settings; the next Wine Steam restart restores `RetinaMode=N` and 96 DPI without deleting or rebuilding the prefix.

### DIFF
--- a/app/src-c/include/metalsharp_backend/config.h
+++ b/app/src-c/include/metalsharp_backend/config.h
@@ -7,5 +7,6 @@
 char* ms_config_get_json(const char* metalsharp_home);
 char* ms_config_set_json(const char* metalsharp_home, const unsigned char* body, size_t body_length, int* status);
 bool ms_config_msync_enabled(const char* metalsharp_home);
+bool ms_config_retina_enabled(const char* metalsharp_home);
 
 #endif
--- a/app/src-c/runtime/config.c
+++ b/app/src-c/runtime/config.c
@@ -198,6 +198,15 @@ bool ms_config_msync_enabled(const char* metalsharp_home) {
     return enabled;
 }
 
+bool ms_config_retina_enabled(const char* metalsharp_home) {
+    char* path = config_path(metalsharp_home);
+    ms_json* config = path == NULL ? NULL : read_json_file(path);
+    bool enabled = config_bool(config, "retinaMode", true);
+    free(path);
+    ms_json_free(config);
+    return enabled;
+}
+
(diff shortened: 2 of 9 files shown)


aaf2tbz:
Very unique pr, I like it.

